PR #11, opened 2026-06-16
Feature/multiarch release bundle

eetoc:
## Checklist

- [ ] Documentation updated when behavior or configuration changed.
- [ ] Tests added or updated for user-visible behavior.
- [ ] No secrets, private endpoints, internal certificates, or local runtime state included.


wuweilceshi-art:
Thanks for working on the multiarch release bundle.

One local data point from macOS arm64 testing:

- Standalone BoxLite CLI v0.9.5 works on macOS arm64:
  `boxlite run alpine:latest echo boxlite-ok`
- `agent-compose-darwin` can normalize a compose config with `driver.boxlite`.
- But starting a session fails with:
  `agent-compose was built without BoxLite cgo support; rebuild with -tags boxlitecgo after preparing libboxlite`
- The current `scripts/export-boxlite-dev-artifact.sh ./build/boxlite` path prepares Linux arm64 ELF artifacts, so they cannot be used directly by the Darwin binary.

For this release bundle work, it would be helpful to clarify the intended support matrix:

1. Does the linux/arm64 bundle include BoxLite cgo support and the required runtime artifacts?
2. Is macOS arm64 BoxLite driver support intended, or should the Darwin binary fail earlier / document that BoxLite driver is Linux-only for now?
3. If macOS support is desired later, would the preferred direction be Darwin cgo artifacts or a CLI/JS-SDK-backed BoxLite driver path?

This is not blocking our current Docker runtime path, but it would help avoid confusion because `driver.boxlite` is accepted by config and only fails at session start on macOS.

eetoc:
> Thanks for working on the multiarch release bundle.
> 
> One local data point from macOS arm64 testing:
> 
> * Standalone BoxLite CLI v0.9.5 works on macOS arm64:
>   `boxlite run alpine:latest echo boxlite-ok`
> * `agent-compose-darwin` can normalize a compose config with `driver.boxlite`.
> * But starting a session fails with:
>   `agent-compose was built without BoxLite cgo support; rebuild with -tags boxlitecgo after preparing libboxlite`
> * The current `scripts/export-boxlite-dev-artifact.sh ./build/boxlite` path prepares Linux arm64 ELF artifacts, so they cannot be used directly by the Darwin binary.
> 
> For this release bundle work, it would be helpful to clarify the intended support matrix:
> 
> 1. Does the linux/arm64 bundle include BoxLite cgo support and the required runtime artifacts?
> 2. Is macOS arm64 BoxLite driver support intended, or should the Darwin binary fail earlier / document that BoxLite driver is Linux-only for now?
> 3. If macOS support is desired later, would the preferred direction be Darwin cgo artifacts or a CLI/JS-SDK-backed BoxLite driver path?
> 
> This is not blocking our current Docker runtime path, but it would help avoid confusion because `driver.boxlite` is accepted by config and only fails at session start on macOS.


Thanks for the detailed macOS arm64 data point.

  For this PR, the intended support matrix is:

  - Linux container images: include the runtime artifacts used by the containerized agent-compose build. BoxLite/Microsandbox paths are intended to work there, subject to the existing host requirements such as /dev/kvm.
  - Go binary release assets: built with CGO_ENABLED=0 and intended for the Docker runtime path only, including macOS arm64/amd64. They do not include BoxLite cgo support or native BoxLite runtime artifacts.

  So macOS arm64 BoxLite driver support is not intended as part of this release bundle. I agree the current behavior is confusing because config normalization accepts driver.boxlite and the failure only appears at session start. I will
  keep the Darwin binary documented as Docker-runtime-only for this PR, and a follow-up should make unsupported native drivers fail earlier or surface a clearer platform/build-capability error.

  If we want macOS BoxLite support later, I think that should be handled separately after deciding between shipping Darwin cgo artifacts or adding a CLI/JS-SDK-backed driver path.

PR #18, opened 2026-06-16
feat: Add DeepSeek provider with configurable default model

TyreeZhao:
## Summary

This PR adds `deepseek` as a first-class guest agent provider in agent-compose.

- Runtime support via DeepSeek OpenAI-compatible Chat Completions API (streaming).
- Backend/loader support to accept `deepseek` and pass `DEEPSEEK_API_KEY` as a guest secret.
- Frontend and SDK provider enums updated.
- Docs updated (`README.md`, `.env.example`).

## Configuration

DeepSeek is configured through global environment variables:

- `DEEPSEEK_API_KEY` (required)
- `DEEPSEEK_MODEL` (optional, default: `deepseek-v4-flash`)
- `DEEPSEEK_API_BASE` (optional, default: `https://api.deepseek.com`)

To change model behavior, set `DEEPSEEK_MODEL` globally (no code change required).

## Notes

- DeepSeek provider is HTTP-based only (no Codex/Claude-style CLI toolchain).
- Rebuild guest image before creating new sessions:
  - `task image:agent-compose-guest`

## Testing

- [x] `cd runtime/javascript && npm run build && npm run test:unit`
- [x] `cd runtime/agent-compose-runtime-sdk && npm test`
- [x] `npm run build:ui`

## Checklist

- [x] Documentation updated for behavior/config changes.
- [x] Tests added/updated for user-visible behavior.
- [x] No secrets or local runtime state included.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,10 @@ HTTP_LISTEN=127.0.0.1:7410
 # OPENAI_API_KEY=
 # LLM_MODEL=
 # LLM_TIMEOUT=60s
+#
+# DeepSeek agent provider uses DEEPSEEK_API_KEY in guest containers (HTTP API, no CLI config).
+# Optional: DEEPSEEK_MODEL=deepseek-v4-flash, DEEPSEEK_API_BASE=https://api.deepseek.com
+# DEEPSEEK_API_KEY=sk-...
 
 # Runtime
 RUNTIME_DRIVER=docker
--- a/README.md
+++ b/README.md
@@ -143,7 +143,8 @@ Top-level fields:
 
 Common agent fields:
 
-- `provider`, `model`, `system_prompt`: agent and LLM settings.
+- `provider`, `model`, `system_prompt`: agent and LLM settings. Supported guest
+  providers are `codex`, `claude`, `gemini`, and `deepseek`.
 - `image`: guest image reference.
 - `driver`: runtime driver override. Supported drivers are `boxlite`, `docker`,
   and `microsandbox`.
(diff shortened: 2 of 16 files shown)


eetoc:
I don't think this should be merged as-is.

  The implementation adds `deepseek` as a first-class agent provider, but the runner is only a single Chat Completions HTTP call. It does not use the workspace, cannot execute tools or mutate files, does not
  support MCP/capability calls, and does not preserve conversation state across turns. That makes it behave very differently from the existing agent providers while being exposed in the same UI/provider path.

  A few concrete issues:

  1. `DeepSeekRunner` ignores the workspace/runtime agent contract. It only sends `messages` to DeepSeek and returns text, so tasks that require file edits or command/tool execution will appear to complete
  without actually changing the workspace.

  2. The returned DeepSeek response `id` is used as `sessionId`, but no provider state is stored or restored. Follow-up messages will not include prior conversation history, so the UI will look like it has an
  agent session while the runtime is stateless.

  3. `outputSchema` is not really enforced. The runner only adds a generic JSON instruction and `response_format: { type: "json_object" }`; it does not send the schema or validate the returned JSON against
  it.

  4. The agent definition `model` field is not wired into the runtime. The runner only reads `DEEPSEEK_MODEL` from the process environment, so model values configured through YAML/UI do not affect DeepSeek
  runs.

  I think DeepSeek support would fit better in the LLM client/protocol layer, or this provider needs to be explicitly modeled as a limited text-only provider rather than a full agent provider.

TyreeZhao:
@eetoc 
Agree with your review — I've reworked this to remove DeepSeek as a guest agent provider and instead add LLM_API_PROTOCOL=chat_completions on the daemon LLMClient for OpenAI-compatible backends (LLMService / scheduler.llm / runtime.llm). Agent providers stay codex / claude / gemini only.

This avoids exposing a stateless HTTP text call on the same UI/path as full workspace agents. OK to update the PR in this direction?

PR #24, opened 2026-06-18
feat: Wire agent system_prompt into guest runtime (Phase 1)

TyreeZhao:
## Summary

- Resolve `AgentDefinition.system_prompt` at run time on the Go host and pass it to the guest via `--system-prompt-file`, with a fallback retry for older guest runtimes that do not recognize the flag.
- Compose Agent Identity + the existing MPI catalog into a single `systemContext` in `agent-compose-runtime-js`, injected into Codex `developer_instructions`, Claude `systemPrompt.append`, and Gemini prompt text as an interim prepend until a native system channel exists.
- Propagate agent definition binding through loader runs and managed project runs, and add host/runtime tests plus EN + zh-CN design/runtime docs.

## Motivation

`system_prompt` was already persisted, exposed in API/Proto, and editable in the Agents UI, but the execution path never read it. Only the MPI catalog reached provider system/developer instruction channels, so configured agent personas had no effect.

### Automated
- [x] `go test ./pkg/agentcompose -run 'SystemPrompt|AgentSystem' -count=1`
- [x] `npm test -- --run` in `runtime/javascript`

### Manual
- [x] Manual: agent with `system_prompt: "Reply only in Chinese"` obeys after Codex/Claude chat run
- [x] Manual: empty `system_prompt` skips `system-prompts/` and agent runs normally (simplified regression; no OctoBus/capsets in local env)
- [x] Manual: Codex session resume after prompt edit uses updated instructions
- [x] Manual: loader bound to agent definition inherits `system_prompt`

### Manual test notes
- Verified on local `go run ./cmd/agent-compose daemon` with data under `~/.local/share/agent-compose/`.
- Empty `system_prompt` regression confirmed by filesystem check: `state/agents/prompts/` present, `system-prompts/` absent.
- Full MPI-only comparison was not exercised locally because OctoBus capability sets are unavailable in this environment.


eetoc:
Strong Phase 1: well tested and documented.

One thing I would reconsider is the `--system-prompt-file` transport. The compatibility retry only seems necessary because system prompt delivery is modeled as a new guest CLI flag. Old guest runtimes do not have that flag registered, so they fail with an unknown option, and the host has to detect that and retry without it.

Could we instead make the runtime discover the system prompt by convention from `--state-root`, similar to how MPI already works?

Today:

```sh
agent-compose-runtime prompt \
  --message-file <p> \
  --system-prompt-file <state>/agents/system-prompts/<provider>-<nano>.txt
```

Alternative:

```text
host writes:
  <state>/agents/system-prompts/system-prompt.txt

runtime reads:
  path.join(stateRoot, "agents", "system-prompts", "system-prompt.txt")
```

`readSystemPromptFile` already degrades to `""` for missing files, so an old/missing prompt state naturally becomes no agent identity without needing a new flag or a retry path.

This would let us delete:

- the `--system-prompt-file` option in `runtime/javascript/src/cli.ts`
- `isUnknownSystemPromptFileFlag`
- the retry block in `executeAgentRun`
- the `--system-prompt-file` append in `buildAgentExecSpec`
- the tests that only cover compatibility retry / flag plumbing

Why I think this is safe:

- The same host state to guest state mapping already exists for `--message-file`.
- MPI already uses convention-based discovery instead of a dedicated CLI flag.
- MPI lives in a separate runtime tree, so this does not affect the existing catalog path.
- A fixed provider-independent filename avoids coupling Go `normalizeAgentKind` to runtime provider normalization.

One implementation detail: if `system_prompt` becomes empty, the host should remove the fixed `system-prompt.txt`; otherwise a later run in the same session could read stale identity text.

One thing to confirm is concurrency. A fixed filename assumes we do not run multiple different agent identities concurrently in the same session. If session agent runs are serialized, this should be fine. If concurrent runs are possible, the current per-run filename avoids overwrite races.

Net: I think convention-based discovery is worth considering because it removes the version-mismatch failure mode instead of patching around it.

eetoc:
Heads up: this PR was merged by mistake and has been reverted in #37.

The main reason is that the current `--system-prompt-file` transport still carries the guest runtime version-mismatch issue discussed above. Since this PR is already merged, it cannot be reopened; the follow-up should be a new PR based on `main`.

For the next attempt, I think we should use convention-based discovery from `--state-root` instead of a new runtime CLI flag, and also handle the empty prompt case by removing the fixed `system-prompt.txt` to avoid stale identity text.

TyreeZhao:
Hi @eetoc — I reworked Phase 1 per your #24 / #37 feedback on fork branch `feat/agent-system-prompt-convention` (single squashed commit). Checking with you before opening a new PR.

**Changes vs #24**

- Dropped `--system-prompt-file`, `isUnknownSystemPromptFileFlag`, and the compatibility retry.
- Host writes `{hostSessionDir}/state/agents/system-prompts/system-prompt.txt`; guest reads it from `--state-root` by convention (MPI-style, no new CLI flag).
- Empty `system_prompt` → delete the file (no stale identity).
- Guest composes Agent Identity + MPI into `systemContext` for Codex / Claude / Gemini.

**Concurrency**

I'm **not** addressing concurrent agent runs in the same session with different identities in Phase 1. A fixed `system-prompt.txt` is overwritten before each exec; if runs overlapped, last write wins — acceptable for now. Current paths are effectively sequential; I documented this assumption.

**Rollout**

Old guests stay MPI-only (no crash). An updated guest runtime is required for agent identity to take effect.

OK for me to open a new PR from `main`? Happy to adjust first.

Thanks!

PR #36, opened 2026-06-21
feat(llm): add chat_completions protocol for OpenAI-compatible backends

TyreeZhao:
## Summary

Follow-up to the closed PR [#18](https://github.com/chaitin/agent-compose/pull/18) (`feat/deepseek-provider`).

In #18, `deepseek` was added as a first-class **guest agent provider**. Review from @eetoc noted that the runner was only a single Chat Completions HTTP call: no workspace access, no tools/MCP, no durable provider session, and misleading parity with `codex` / `claude` / `gemini` on the same UI/provider path.

This PR implements the recommended direction from that review: **move OpenAI-compatible Chat Completions support to the daemon LLM client/protocol layer**, instead of exposing it as a guest agent provider.

### What changed

- Add **`LLM_API_PROTOCOL`** daemon configuration:
  - `responses` (default, existing OpenAI Responses API behavior)
  - `chat_completions` for OpenAI-compatible Chat Completions backends
  - aliases: `chat`, `chat_completion`
- Extend **`LLMClient`** with protocol-aware routing and endpoint normalization (`/v1/chat/completions`)
- Align LLM config resolution: **global env → process env → daemon config**
- Unify API key lookup order: `LLM_API_KEY`, then `OPENAI_API_KEY`
- Add tests for protocol resolution, chat completions success/error/schema paths, `LLMService.Generate`, and loader host integration
- Update docs (English/Chinese README, design docs, loader-script, AGENTS.md, SDK README, `.env.example`)

### What this PR does *not* do

- Does **not** add `deepseek` as a guest agent provider
- Does **not** change guest runtime runners, frontend provider enums, or agent session semantics
- Does **not** create workspace-capable agent sessions via Chat Completions

Guest agent providers remain **`codex` / `claude` / `gemini` only**.

### How to use DeepSeek (and similar backends)

Configure daemon-side LLM settings (`.env` or UI global env):

```env
LLM_API_PROTOCOL=chat_completions
LLM_API_ENDPOINT=https://api.deepseek.com
LLM_API_KEY=...
LLM_MODEL=deepseek-v4-flash
```

Used by:

- `LLMService.Generate`
- loader `scheduler.llm`
- SDK `runtime.llm`

This path is for **unary text generation only** — no file, command, or MCP tool access.

### Architecture

| Capability | Path | Workspace / tools |
| --- | --- | --- |
| Agent | guest `agent-compose-runtime prompt` | yes |
| LLM | daemon `LLMClient` | no |

## Testing

```bash
go test ./pkg/agentcompose/ -run 'LLM|NormalizeLLM|LoaderRunHost|ResolveSetting|ServiceGenerateLLM' -count=1
go test ./pkg/config/ -count=1
golangci-lint run --allow-parallel-runners ./pkg/agentcompose/... ./pkg/config/...
```

Manual verification:

```bash
# Start daemon with chat_completions config, then:
curl -sS -X POST 'http://127.0.0.1:7410/agentcompose.v1.LLMService/Generate' \
  -H 'Content-Type: application/json' \
  -d '{"prompt":"Say hello in one sentence","model":"deepseek-v4-flash"}'
```

Verified locally against an OpenAI-compatible Chat Completions backend (upstream errors correctly surfaced when model name was invalid).

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,10 +26,19 @@ HTTP_LISTEN=127.0.0.1:7410
 
 # LLM
 # LLM_API_ENDPOINT=https://api.openai.com
+# LLM_API_PROTOCOL=responses
+# Accepted aliases for chat_completions: chat, chat_completion
 # LLM_API_KEY=
 # OPENAI_API_KEY=
 # LLM_MODEL=
 # LLM_TIMEOUT=60s
+#
+# OpenAI-compatible chat completions (e.g. DeepSeek, local vLLM/Ollama).
+# Use chat_completions instead of responses (do not set both):
+# LLM_API_PROTOCOL=chat_completions
+# LLM_API_ENDPOINT=https://api.example.com
+# LLM_API_KEY=...
+# LLM_MODEL=your-model
 
 # Runtime
 RUNTIME_DRIVER=docker
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,6 +55,11 @@ Important defaults:
 - `DEFAULT_IMAGE`: `debian:bookworm-slim`
 - `JUPYTER_PROXY_BASE`: `/jupyter`
 
+Daemon LLM client (`LLMService`, `scheduler.llm`, SDK `runtime.llm`):
+- `LLM_API_ENDPOINT`, `LLM_API_KEY`, `OPENAI_API_KEY`, `LLM_MODEL`, `LLM_TIMEOUT`
+- `LLM_API_PROTOCOL`: `responses` (default) or `chat_completions` for OpenAI-compatible backends (aliases: `chat`, `chat_completion`)
+- Guest agent providers remain `codex`, `claude`, and `gemini` CLI runners in guest containers
+
 ## Persistence
 
 Session metadata, notebook cells, event history, runtime state, and proxy state are stored under `SESSION_ROOT`.
--- a/README.md
+++ b/README.md
@@ -143,7 +143,10 @@ Top-level fields:
 
 Common agent fields:
 
-- `provider`, `model`, `system_prompt`: agent and LLM settings.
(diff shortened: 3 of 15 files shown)


TyreeZhao:
image

eetoc:
LGTM. I think this is the right direction and safe to merge.

This keeps the architecture boundary clear: OpenAI-compatible Chat Completions extends the daemon LLM path (`LLMService.Generate` / `scheduler.llm`) instead of pretending to be a guest agent provider. Guest agent providers and workspace/tool/session semantics remain unchanged.

I also verified the main path with `LLM_API_PROTOCOL=chat_completions` against the real DeepSeek endpoint:

- `LLM_API_ENDPOINT=https://api.deepseek.com`
- `LLM_MODEL=deepseek-v4-flash`
- `LLMService.Generate` works for normal text generation
- `LLMService.Generate` also works with `outputSchema` and returns JSON

Since `responses` remains the default when `LLM_API_PROTOCOL` is unset, this looks like a backward-compatible extension of the existing daemon LLM capability.

One non-blocking follow-up: for `chat_completions`, `outputSchema` uses prompt guidance + `response_format=json_object`, so it is not equivalent to Responses API strict schema enforcement. This can be documented separately.

eetoc:
One more non-blocking follow-up thought: `LLM_API_PROTOCOL` is a good first step and keeps this PR backward-compatible, but longer term we may want per-call selection among configured daemon LLM backends.

For example, the daemon could define named LLM backends such as `default` and `deepseek`, while `LLMService.Generate` / `scheduler.llm` can select one by name. The API key and endpoint would still stay in daemon/global config, not in request parameters.

That would allow one deployment to use multiple daemon-side LLM backends without mixing this with guest agent providers. Not needed for this PR, but it feels like a natural next step after protocol support lands.

PR #61, opened 2026-06-22
feat: Wire agent system_prompt via convention path discovery (Phase 1)

TyreeZhao:
Follow-up to [#24](https://github.com/chaitin/agent-compose/pull/24) / [#37](https://github.com/chaitin/agent-compose/pull/37). Per @eetoc review: convention-based transport — no new guest CLI flag, no host retry.

## Summary

- Host resolves `AgentDefinition.system_prompt` at run time and writes/removes `<session>/state/agents/system-prompts/system-prompt.txt`
- Guest reads it from `--state-root`, composes Agent Identity + MPI into `systemContext`, injects into Codex / Claude / Gemini
- Loader runs and managed project runs propagate agent definition binding
- Empty `system_prompt` deletes the file to avoid stale identity

## vs #24

Dropped `--system-prompt-file`, `isUnknownSystemPromptFileFlag`, and compatibility retry. Fixed provider-independent filename under `--state-root` (MPI-style discovery).

## Notes

- **Old guest runtime**: MPI-only, no crash
- **Updated guest runtime**: required for agent identity to take effect
- **Concurrency**: fixed filename assumes serialized agent runs per session (Phase 1)

### Automated

- [x] `go test ./pkg/agentcompose -run 'SystemPrompt|AgentSystem' -count=1`
- [x] `npm test -- --run` in `runtime/javascript`

### Manual

- [x] Create a Codex or Claude agent with `system_prompt: "Reply only in Chinese"`, run chat → replies in Chinese
- [x] Clear `system_prompt`, run again → MPI-only, normal completion; `system-prompt.txt` absent
- [x] Edit `system_prompt` on the same Codex session, send another message → new instructions apply

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,6 +18,7 @@ The design notes below describe current implementation details. Chinese versions
 are available under [zh-CN/design/](zh-CN/design/).
 
 - [Architecture notes](design/agent-compose_design.md)
+- [Agent system prompt (Phase 1)](design/agent_system_prompt_design.md)
 - [Runtime JavaScript contract](design/agent-compose-runtime-js_contract.md)
 - [Runtime environment variables](design/runtime_environment_variables_design.md)
 - [Runtime mount manifest](design/runtime_mount_manifest_design.md)
--- a/docs/design/agent-compose-runtime-js_contract.md
+++ b/docs/design/agent-compose-runtime-js_contract.md
@@ -121,6 +121,20 @@ guest: /data/state/agents/prompts/<provider>-<unix_nano>.txt
 
 The guest path is then passed to the JavaScript runtime through `--message-file`.
 
+When a run is bound to an agent definition with non-empty `system_prompt`, the
+host writes the trimmed text to a fixed convention path:
+
+```text
+host:  <session>/state/agents/system-prompts/system-prompt.txt
+guest: /data/state/agents/system-prompts/system-prompt.txt
+```
+
+The guest runtime reads this path via `agentSystemPromptPath(stateRoot)` in
+`prompt.ts`. If the file is missing or empty, `readSystemPromptFile` returns
+`""` and the run composes MPI-only context. When `system_prompt` becomes empty,
+the host removes `system-prompt.txt` to avoid stale identity on later runs in
+the same session.
+
 ### 3.3 Agent HOME And Initial Config
 
 The host sets these values for agent execution:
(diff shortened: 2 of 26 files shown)


eetoc:
I tested this branch with a real Claude-compatible backend.

Environment used:

- `CLAUDE_CODE_EXECUTABLE`: Claude Code 2.1.159
- `@anthropic-ai/claude-agent-sdk`: 0.2.141
- backend: Claude-compatible endpoint
- model: `kimi-k2.6`

### Agent Identity path

I created a temporary runtime state with:

- `state/agents/system-prompts/system-prompt.txt`
- provider: `claude`
- prompt message: `hello`

The system prompt file contained:

```text
You must reply with exactly this text and nothing else: AC_SYSTEM_PROMPT_HELLO_OK
```

The runtime command completed successfully and returned:

```json
{
  "provider": "claude",
  "stopReason": "end_turn",
  "finalText": "AC_SYSTEM_PROMPT_HELLO_OK"
}
```

So the convention path works for Claude: the runner reads the file from `--state-root`, builds `systemContext`, and passes it through to Claude.

### MPI coexistence

I also tested with both files present:

- `state/agents/system-prompts/system-prompt.txt`
- `runtime/mpi/catalog.md`

The system prompt file contained:

```text
You are testing coexistence with MPI. Follow the MPI catalog if it contains a verification phrase.
```

The MPI catalog contained:

```markdown
# MPI Verification Catalog

For this verification, when the user says hello, reply with exactly this text and nothing else: AC_MPI_CATALOG_OK
```

The runtime returned:

```json
{
  "provider": "claude",
  "stopReason": "end_turn",
  "finalText": "AC_MPI_CATALOG_OK"
}
```

This verifies that the `mpiContext` -> `systemContext` change does not appear to drop MPI in the Claude path. Claude received both Agent Identity and the MPI catalog, and the MPI catalog remained effective.

I also checked the local provider wiring shape: `ClaudeRunner.queryOptions().systemPrompt.append` contains both `## Agent Identity` and `## MPI Catalog`, with Agent Identity before MPI.

### Resume semantics

One semantic point worth clarifying: resumed provider sessions may continue to follow the system prompt they were originally created with. I think that behavior is reasonable for provider thread continuity, but the PR text saying empty prompts remove the file “to avoid stale identity” could be read as meaning edits/clears affect an already-resumed provider session.

Suggested clarification:

> `system_prompt` is applied when starting a provider session. Resumed provider sessions may retain the identity they were created with; clearing the file prevents new provider sessions from reading stale identity, but does not rewrite already-resumed provider threads.

### Non-blocking test suggestion

The current pure `buildSystemContext` coverage is useful, and the runtime tests pass. It would still be good to add a provider-level regression test asserting that the final provider injection contains both `## Agent Identity` and `## MPI Catalog` for Claude/Codex/Gemini. That would guard the provider-specific wiring after the `mpiContext` -> `systemContext` change.

LGTM from my side after the resume semantics wording is clarified.

PR #67, opened 2026-06-22
feat: add OpenCode provider support

eetoc:
## Summary
- add `opencode` provider normalization, validation, UI selection, SDK/docs coverage
- add OpenCode CLI runner in `agent-compose-runtime-js` with model/system prompt plumbing and session state reuse
- install and link `opencode-ai` in the guest image

## Tests
- `npm run typecheck && npm run test:unit` in `runtime/javascript`
- `task test`
- `GOFLAGS=-buildvcs=false task lint`
- `GOFLAGS=-buildvcs=false task build`
- `git diff --check`

Note: plain Go build/lint VCS stamping fails in this `/tmp` worktree with `error obtaining VCS status`; the same gates pass with `GOFLAGS=-buildvcs=false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,8 +147,9 @@ Top-level fields:
 Common agent fields:
 
 - `provider`, `model`, `system_prompt`: guest agent settings (`provider` selects
-  the guest CLI runner; `model` is passed to that agent runtime). Supported guest
-  providers are `codex`, `claude`, and `gemini`. Daemon-side LLM calls
+  the guest CLI runner; `model` is passed to provider runtimes that support
+  explicit model selection). Supported guest providers are `codex`, `claude`,
+  `gemini`, and `opencode`. Daemon-side LLM calls
   (`LLMService`, `scheduler.llm`) use `LLM_MODEL` instead.
 - `image`: guest image reference.
 - `driver`: runtime driver override. Supported drivers are `boxlite`, `docker`,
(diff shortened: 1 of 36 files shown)


eetoc:
I took another pass over this PR. The branch itself still looks internally healthy from the focused checks I ran:

```bash
npm run typecheck
npm run test:unit
go test ./pkg/agentcompose -run 'OpenCode|AgentSessionMessageUsesDefinitionProvider|AgentDefinition|Loader|RunAgent' -count=1
git diff --check origin/main...HEAD
```

Those pass. I also checked the current `opencode-ai` npm metadata, and the package bin is still `opencode: bin/opencode.exe`, so the guest image symlink target looks reasonable.

The main blocker now is that the PR is conflicting with current `main` (`mergeable: CONFLICTING`). The important conflict area is not just docs: current `main` has since added the agent system prompt convention and runtime LLM facade path. When rebasing, I think the OpenCode integration should be adapted to those mainline paths instead of keeping the older parallel wiring from this branch.

In particular:

- Current `main` writes agent identity to the fixed convention path `state/agents/system-prompts/system-prompt.txt`, and the JS runtime discovers it from `--state-root` via `system-context.ts`.
- This PR currently adds a separate `--system-prompt-file` CLI option and per-run files like `state/agents/system-prompts/opencode-*.txt`.
- After rebase, OpenCode should probably use the same `buildSystemContext(...)` / fixed convention path model as Codex and Claude, rather than carrying a separate system prompt flag/path.

The other rebase-sensitive area is `ExecuteAgentRequest` / `executeAgentRun`: current `main` carries `AgentDefinitionID`, `Model`, and `RunID`, and calls `ensureSessionLLMFacadeConfig(...)` before execution. Please make sure the OpenCode model plumbing is layered onto that current flow instead of replacing it with the older branch shape.

Once those conflicts are resolved against current `main`, I think this should be much easier to review for the actual OpenCode runner behavior.

eetoc:
Real regression testing completed with the actual LLM configuration:

- `LLM_API_ENDPOINT` / `ANTHROPIC_BASE_URL`: `https://aiapi.chaitin.net`
- `LLM_MODEL` / `ANTHROPIC_MODEL`: `kimi-k2.6`
- `RUNTIME_DRIVER`: `docker`
- Guest image rebuilt: `agent-compose-guest:latest`

Tested flows:

1. OpenCode custom provider through the agent-compose LLM facade
   - Created an `opencode` agent with model `chaitin/kimi-k2.6`.
   - Created a Docker runtime session using `agent-compose-guest:latest`.
   - Sent a real agent message through `AgentService/SendAgentMessage`.
   - The OpenCode CLI called the runtime facade from inside the Docker container.
   - The request completed successfully and returned `opencode-model-fetch-disabled-ok`.

2. OpenCode runtime config mounting
   - Verified the generated OpenCode config was written under the session home.
   - Verified Docker mounts `.opencode` into `/root/.opencode`.
   - Verified `OPENCODE_CONFIG` points to `/root/.opencode/agent-compose.json`.
   - Verified the container can resolve the custom provider/model `chaitin/kimi-k2.6`.

3. OpenCode model catalog fetch behavior
   - Rebuilt the guest image with `OPENCODE_DISABLE_MODELS_FETCH=1`.
   - Verified OpenCode no longer creates `/root/.cache/opencode/models.json` at runtime.
   - Verified the host session home also does not contain `models.json` after the real agent run.

4. Daemon LLM service sanity check
   - Called `LLMService.Generate` with model `kimi-k2.6`.
   - The request completed successfully and returned `llm-generate-ok`.

Cleanup:

- Stopped the Docker runtime session.
- Stopped the temporary daemon.
- Confirmed no `agent-compose` test containers were left running.

Merge assessment: from the regression results above and the focused unit coverage, this PR looks ready to merge from my side.

PR #210, opened 2026-07-07
set microsandbox bind mount quota

eetoc:
## Summary
- add MICROSANDBOX_BIND_QUOTA_GB for microsandbox bind mount guest-write quota
- default the bind quota to BOX_DISK_SIZE_GB so existing 60G deployments also get 60G /data headroom
- explicitly set QuotaMiB on microsandbox bind mounts to avoid upstream's 4GiB default

## Root Cause
Microsandbox v0.5.10+ applies a default 4GiB guest-write quota to bind mounts. Agent Compose did not pass an explicit quota for the directory-only /data bind mount, so /data was limited to baseline + 4GiB even when BOX_DISK_SIZE_GB configured the docker disk to 60G.

## Validation
- go test ./pkg/config ./pkg/driver

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -103,6 +103,9 @@ DEFAULT_IMAGE=ghcr.io/chaitin/agent-compose-guest:latest
 # Docker and MicroSandbox storage.
 # --------------------------------------
 # DOCKER_HOME=/data/docker
+# Microsandbox bind mounts get this guest-write quota. Defaults to
+# BOX_DISK_SIZE_GB when unset.
+# MICROSANDBOX_BIND_QUOTA_GB=6
 # MICROSANDBOX_INSECURE_REGISTRIES=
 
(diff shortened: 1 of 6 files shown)


eetoc:
Validation completed.

Checks:
- `go test ./pkg/config ./pkg/driver`

Runtime verification:
- Verified microsandbox bind mount with an explicit 60 GiB quota reports `/data` as 60 GiB instead of the upstream 4 GiB default.
- Verified writing 5 GiB to `/data` succeeds with `DD_RC=0`.
- Verified the mount reports approximately 55 GiB available after the 5 GiB write.

This confirms the PR behavior: when Agent Compose sets `QuotaMiB` on the microsandbox bind mount, `/data` no longer falls back to the default 4 GiB bind quota.

PR #226, opened 2026-07-08
Improve logs input output display

eetoc:
## Summary
- show USER and ASSISTANT sections in text logs by default
- include prompt in logs JSON while keeping content as output
- handle delayed prompts for follow streams, including command runs with no output

## Validation
- go test ./cmd/agent-compose
- docker build agent-compose:latest with Dockerfile.agent-compose-dockeronly
- docker compose recreate agent-compose
- container CLI follow logs validation for command runs with and without output

image


eetoc:
@monkeyscan 整体review下这个mr

PR #278, opened 2026-07-13
fix: bound microsandbox exec stream drain after process exit

eetoc:
## Summary

- bound microsandbox ExecStream draining after ExecEventExited with a 2-second idle grace period
- close lingering exec handles and return the recorded exit code when the relay never emits ExecEventDone
- add a regression test for Exited without Done

## Validation

- `go test ./pkg/driver -count=1`
- `go test ./pkg/driver -run ^TestMicrosandbox -count=1`
- `git diff --check`

The full suite still has pre-existing macOS `/private/var` path assertion failures in `pkg/volumes`; Go lint is blocked by the pre-existing unused `boxliteVolumeBridgeIsMountPoint` function.

### DIFF
--- a/pkg/driver/microsandbox_runtime.go
+++ b/pkg/driver/microsandbox_runtime.go
@@ -42,6 +42,99 @@ type microsandboxExecCollector struct {
 	output bytes.Buffer
 }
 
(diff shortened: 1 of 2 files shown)


eetoc:
@monkeyscan review这个MR

PR #294, opened 2026-07-14
fix(cli): print usable Jupyter entry URL

eetoc:
## Summary

- print the Jupyter token-redirect entry when a run enables Jupyter without exposing its tokenized notebook URL
- preserve the direct `/lab?token=...` URL when the daemon returns an exposed notebook URL
- keep the reported proxy path unchanged for JSON consumers
- add regression coverage for non-exposed, exposed, trailing-slash, and already-normalized paths

## Verification

- `go test ./cmd/agent-compose -run 'TestIntegrationCLIRun(JupyterWithoutExposePrintsTokenRedirectEntry|SendsJupyterExpose)|TestJupyterBrowserLocation' -count=1`\n- `go test ./cmd/agent-compose -count=1`\n- `task lint`\n- real Docker sandbox: the non-exposed URL redirected to a tokenized JupyterLab URL and completed with HTTP 200\n- real Docker sandbox: the exposed tokenized URL opened JupyterLab directly with HTTP 200\n\n## Known environment failures\n\n- `task test` reached `deploy/install_test.sh` and failed an existing incomplete-rollback message assertion (`recovery was incomplete`)\n- the full Linux artifact build could not refresh runtime artifacts because Docker Hub timed out while fetching `golang:1-alpine`

### DIFF
--- a/cmd/agent-compose/cli_jupyter.go
+++ b/cmd/agent-compose/cli_jupyter.go
@@ -54,9 +54,21 @@ func resolveRunJupyterOutput(ctx context.Context, cli cliOptions, sandboxID stri
 		return composeRunJupyterOutput{}, err
 	}
 	return composeRunJupyterOutput{
-		URL:  joinBaseURLAndPath(baseURL, firstNonEmptyString(notebookURL, proxyPath)),
+		URL:  joinBaseURLAndPath(baseURL, jupyterBrowserLocation(notebookURL, proxyPath)),
 		Path: proxyPath,
 	}, nil
+}
+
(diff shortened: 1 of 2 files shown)


eetoc:
Manual verification completed with real Docker sandboxes:

- `--jupyter`: CLI printed the sandbox entry URL without `/lab` or a token; the entry returned HTTP 307 to the tokenized JupyterLab URL, which returned HTTP 200.
- `--jupyter-expose`: CLI printed the tokenized `/lab` URL directly, and it returned HTTP 200 with JupyterLab.

The test sandboxes remained running during browser verification.

PR #299, opened 2026-07-15
fix(runs): index capability tokens for project sandboxes

xiaoss:
## Summary
- index project-run sandboxes with the capability token resolver after the sandbox is running and reloaded
- add regression coverage for both created and reused project-run sandboxes

## Why
Project runs currently receive CAP_TOKEN/CAP_GRPC_TARGET, but the capability proxy cannot resolve the sandbox token because this run path never indexes the sandbox. Capability calls then fail with `Unauthenticated desc = capability sandbox token not found`.

## Testing
- go test ./pkg/runs
- verified against a patched v2607.6.6 daemon that project-run grpcurl can list a configured capset service successfully

### DIFF
--- a/pkg/runs/controller.go
+++ b/pkg/runs/controller.go
@@ -2148,7 +2148,14 @@ func (c *Controller) startProjectRunSandbox(ctx context.Context, sandbox *domain
 	}
 	domain.RestoreSandboxTransientFields(loaded, sandbox)
 	*sandbox = *loaded
+	c.indexCapabilitySandbox(sandbox)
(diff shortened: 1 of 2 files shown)


eetoc:
感谢你提交这个修复并补充相关测试！这个 PR 与 #302 处理的是同一个问题，我们综合考虑生命周期语义和测试覆盖后，已经合并了 #302，因此这里不再重复合并。你的排查和实现对定位、确认这个问题也很有帮助，再次感谢你的贡献！

PR #316, opened 2026-07-16
fix(cli): show scheduler first in list output

kingfs:
## Summary

  - Move the `SCHEDULER` column to the first position in `scheduler ls` text output.
  - Keep the same column order for both default and `--verbose` views.
  - Add regression coverage for the header and row ordering in both views.

  ## Testing

  - `go test ./cmd/agent-compose -run '^TestIntegrationCLISchedulerList$'`
  - `go test ./cmd/agent-compose`
  - `task lint`
  - `task build`
  - `task test`

  ## Checklist

  - [ ] Documentation updated when behavior or configuration changed.


innomentats (approved):
LGTM

PR #327, opened 2026-07-17
feat(microsandbox): inherit Docker image ENV as baseline env vars in guest VM

eetoc:
## Summary

When a Docker image defines `ENV` directives (e.g. `ENV DEBIAN_FRONTEND=noninteractive`), those variables were silently discarded when starting a microsandbox guest. The `createSandbox` path only used user-provided `EnvItems` and hardcoded agent-compose defaults. Docker native runtime inherits image ENV automatically through the daemon, but microsandbox boots from a raw rootfs and must explicitly pass them.

## Changes

- **imagecache**: store `Env []string` in `ImageMetadata`, `MetadataInput`, `MaterializationResult`; extract from `configFile.Config.Env` during pull
- **local_docker_oci**: persist image ENV alongside rootfs cache (`.env.json`) and thread it through `localDockerImageRootfs.Env`
- **microsandbox_image_resolver**: add `Env` to `microsandboxRootFSResult`, propagate from both Docker-daemon and OCI-pull paths
- **microsandbox_runtime**: read image ENV as baseline in `createSandbox`; user `EnvItems` / `RuntimeEnvItems` / agent-compose defaults override (matching Docker daemon behavior)
- **types**: add `parseEnvEntry` helper for `KEY=VALUE` parsing with unit test

## Merge priority

image ENV (baseline) < user EnvItems/RuntimeEnvItems < agent-compose hardcoded defaults

LLM provider keys (`OPENAI_API_KEY`, etc.) are still filtered from image ENV.

## Files changed

12 files, +139 / -15 lines.

### DIFF
--- a/pkg/driver/local_docker_oci.go
+++ b/pkg/driver/local_docker_oci.go
@@ -28,6 +28,7 @@ type localDockerImageRootfs struct {
 	ImageID     string
 	ResolvedRef string
 	RootfsPath  string
+	Env         []string
 }
 
 type localDockerImageLayout struct {
@@ -76,6 +77,8 @@ func materializeLocalDockerImageRootfsWithClient(ctx context.Context, dataRoot s
 		return localDockerImageRootfs{}, false, nil
 	}
 
+	imageEnv := inspectEnv(inspect)
+
 	imageID := strings.TrimPrefix(inspect.ID, "sha256:")
 	if imageID == "" {
 		imageID = sanitizeRuntimeName(resolvedRef)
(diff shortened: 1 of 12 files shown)


eetoc:
Good catch. The Warn logging has been added in e0872f4 so the fallback is no longer silent.

One thing worth noting: when `.env.json` exists but is corrupt (rather than missing), falling back to `fresh` would also be problematic. The cache file may have been written by a previous agent-compose process with a different image state, while `fresh` comes from the current `ImageInspect` — using `fresh` would overwrite that persisted state and potentially diverge from what other processes see. So the current behavior (log + return nil, skipping image ENV for this single sandbox creation) is the safer choice.

innomentats (approved):
LGTM

PR #334, opened 2026-07-17
fix(docs): exclude v1 migration fields from schema coverage

eetoc:
## Summary

- exclude temporary `AgentSpec.display_name`, `AgentSpec.description`, scheduler presentation metadata, and `WorkspaceSpec.commit` from public YAML schema coverage
- scope compatibility exclusions by Go struct and YAML field so future fields with the same YAML name still require documentation
- add a focused regression test for the schema coverage filter

## Testing

- `npm test` (in `tools/genpages`)
- `task docs:build --force`
- `git diff --check`

## Checklist

- [x] Documentation coverage remains aligned with the public v2 authoring surface.
- [x] Tests added or updated for changed behavior.
- [x] No generated `build/pages` files committed.
- [x] No secrets, private endpoints, or local runtime state included.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -81,6 +81,8 @@ tasks:
       - '../../pkg/compose/spec.go'
       - 'check-pages.mjs'
       - 'render-pages.mjs'
+      - 'schema-coverage.mjs'
+      - 'schema-coverage.test.mjs'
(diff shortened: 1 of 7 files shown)


innomentats (approved):
LGTM

PR #339, opened 2026-07-20
fix(microsandbox): isolate guest etc state

eetoc:
## Summary

- copy the materialized image `/etc` into sandbox-owned state before creating a Microsandbox
- mount the sandbox-owned copy at `/etc` so concurrent guests cannot overwrite each other's hostname and resolver configuration through the shared bind rootfs
- preserve the isolated state across stop/resume and publish it atomically

## Testing

- `go test ./pkg/driver`
- `go test -race ./pkg/driver`
- `task test:unit`
- `task build`
- `git diff --check`
- `task lint` was attempted but the installed golangci-lint was built with Go 1.25, below the repository target Go 1.26.2
- `task test` was attempted; deterministic script checks passed, then the deployment installer test failed on macOS because BSD `realpath` does not support `-m`

## Checklist

- [x] Documentation updated when behavior or configuration changed. No public behavior or configuration documentation is affected.

### DIFF
--- a/pkg/driver/microsandbox_etc.go
+++ b/pkg/driver/microsandbox_etc.go
@@ -1,0 +1,113 @@
+package driver
+
+import (
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+const microsandboxEtcStatePath = "microsandbox-rootfs/etc"
+
+// prepareMicrosandboxEtc copies the image's /etc into sandbox-owned state.
+// Microsandbox bind rootfs directories are shared by every sandbox using the
+// same materialized image. Its guest init writes instance-specific hosts and
+// resolver configuration under /etc, so /etc must not remain in that shared
+// directory.
+func prepareMicrosandboxEtc(rootfsPath, sandboxStateDir string) (string, error) {
+	source := filepath.Join(rootfsPath, "etc")
+	info, err := os.Stat(source)
+	if err != nil {
+		return "", fmt.Errorf("stat microsandbox image /etc: %w", err)
+	}
+	if !info.IsDir() {
+		return "", fmt.Errorf("microsandbox image /etc is not a directory")
+	}
+
+	target := filepath.Join(sandboxStateDir, filepath.FromSlash(microsandboxEtcStatePath))
+	if _, err := os.Stat(target); err == nil {
+		return target, nil
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("stat sandbox-owned microsandbox /etc: %w", err)
+	}
+
+	parent := filepath.Dir(target)
+	if err := os.MkdirAll(parent, 0o755); err != nil {
+		return "", fmt.Errorf("create sandbox-owned microsandbox rootfs state: %w", err)
+	}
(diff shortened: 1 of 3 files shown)


eetoc:
### Investigation: why the bug happens

Root cause is the interaction of two facts — a **shared** rootfs and a guest that **unconditionally rewrites `/etc` on every boot**.

**1. The materialized image rootfs is shared, not per-sandbox.**
On the directory-only path, `resolveMicrosandboxImageRef` returns the materialized image rootfs directory (`RootFSPath`, an absolute path) and that single directory is virtiofs-mounted as the guest `/` for **every** sandbox using the same image (`.rootfs.ready` materializes it once, then it is reused across sandboxes). So any write the guest makes to a path under `/` — unless that path is separately shadowed by a tmpfs/disk/bind mount — is written straight back into the shared host directory and persists there.

That is exactly why `/run` (tmpfs) and `/var/lib/docker` (own ext4 disk) are already shadowed. `/etc` was **not**, so it stayed pointed at the shared directory.

**2. Microsandbox's guest init (`agentd`) rewrites instance-specific files under `/etc` on every boot.**
Verified in upstream `superradcompany/microsandbox`, `crates/agentd/lib/network.rs` (@ `e2c8cee`):

- [`set_hostname`](https://github.com/superradcompany/microsandbox/blob/e2c8ceeadaf7f1d0fec8d70ad203024862cfb465/crates/agentd/lib/network.rs#L534) → `fs::write("/etc/hostname", …)` — per-sandbox hostname
- [`write_hosts_file`](https://github.com/superradcompany/microsandbox/blob/e2c8ceeadaf7f1d0fec8d70ad203024862cfb465/crates/agentd/lib/network.rs#L547) → `fs::write("/etc/hosts", …)` — contents include this VM's gateway IPv4/IPv6
- [`write_resolv_conf`](https://github.com/superradcompany/microsandbox/blob/e2c8ceeadaf7f1d0fec8d70ad203024862cfb465/crates/agentd/lib/network.rs#L569) → `fs::write("/etc/resolv.conf", …)` — this VM's DNS servers (called as step 8 of [`configure_interface`](https://github.com/superradcompany/microsandbox/blob/e2c8ceeadaf7f1d0fec8d70ad203024862cfb465/crates/agentd/lib/network.rs#L184))

These are plain, unconditional `fs::write` calls (no overlay / copy-on-write), and every value is **instance-specific** (hostname, gateway IP, DNS).

**Putting them together:** each guest boot writes its own hostname / hosts / resolv.conf into the *shared* image `/etc`. With concurrent (or even just sequential-reuse) sandboxes on the same image, the last writer wins and the others silently get another sandbox's hostname and resolver config through the shared bind rootfs — cross-talk, not isolation.

**Why copy `/etc` instead of a tmpfs (like `/run`):** `/etc` must retain the image's own contents (`passwd`, `nsswitch.conf`, TLS/apt config, …), so an empty tmpfs would break the guest. A per-sandbox **copy**, bind-mounted at `/etc`, gives each guest a private, writable `/etc` that starts from the image and absorbs `agentd`'s per-boot writes without touching the shared rootfs. This mirrors the existing treatment of `/root`, which had the same shared-rootfs pollution class of bug.

innomentats (approved):
LGTM

PR #352, opened 2026-07-21
fix(installer): drop --progress flag and make cancellation reach compose

eetoc:
Two installer failures that only occur on error paths, so neither was covered by existing tests. Both affect current users and are independent of any feature work.

## 1. `--progress plain` breaks pull on older Compose

`Service.compose` unconditionally appended `--progress plain` to `pull` and `up`. That flag only exists on newer Compose plugins. On older ones every install and upgrade dies at the pull step:

```
run docker compose --progress plain pull: exit status 1
unknown flag: --progress
```

Because pull happens inside the install transaction, this then triggers a full rollback — the user sees `pull deployment images: exit status 1` and an installation that reverted itself. Reproduced on Compose **v2.10.2** (Docker Desktop, Server 20.10.17).

`checkCompose` only verifies `docker compose version --short` succeeds, i.e. that Compose v2 exists at all — it never established a version floor, so the incompatibility slipped past the preflight check and surfaced mid-transaction.

The flag is now dropped rather than version-gated: it existed to keep the TUI from dealing with docker's cursor-driven redraws, which is a display concern better solved in the TUI than by constraining which Compose versions the installer supports.

## 2. Cancellation never reaches the process doing the work

`ExecRunner.Run` relied on `exec.CommandContext`, which signals **only the direct child**. The real process tree is three deep:

```
installer
 └─ docker                 ← the only process CommandContext signals
     └─ com.docker.cli
         └─ docker-compose ← actually performing the pull
```

The grandchildren survive and, critically, **inherit the stdout/stderr pipe**. `cmd.Run()` must wait for its output-copying goroutine, which only ends at EOF — and EOF never arrives while a descendant holds the write end. So Ctrl+C in the TUI printed "正在取消并回滚..." and then hung indefinitely.

Fix: run the child in its own process group and signal the whole group on cancel, plus a `WaitDelay` backstop so a process that ignores the signal costs a bounded wait instead of a hang.

Repeated Ctrl+C also appended a duplicate notice per press; it is now recorded once.

## Tests

`TestExecRunnerCancelStopsGrandchildHoldingOutput` reproduces the hang with a shell that backgrounds a child inheriting stdout. Verified to fail before the fix (hit the 8s timeout) and pass after (**0.2s**).

The Compose change is covered by the existing exact-command assertions in `TestE2EInstallerLifecycle`, updated to expect no `--progress`.

Full suite passes, `go vet` clean, `gofmt` clean.

> Note for reviewers on macOS: `internal/core` integration tests fail under the default `TMPDIR` because `/var` is a symlink and `validateInstallPath` rejects symlinked components. This is pre-existing and reproduces on an unmodified tree. Run with `TMPDIR` set to a non-symlinked path; Linux CI is unaffected.

### DIFF
--- a/cmd/installer/internal/core/command.go
+++ b/cmd/installer/internal/core/command.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"strings"
-	"syscall"
 	"time"
 )
 
@@ -32,12 +31,11 @@ func (r ExecRunner) Run(ctx context.Context, dir, name string, args ...string) e
 	// `docker` runs `compose` as a plugin in a separate process that inherits
 	// this output pipe. CommandContext only signals the direct child, so the
 	// plugin would keep running and holding the pipe open, and Wait would block
-	// on it forever. Giving the child its own process group lets cancellation
-	// reach the whole tree.
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
-	cmd.Cancel = func() error { return syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM) }
-	// Backstop for anything that ignores the signal: abandon its output rather
-	// than hang the installer. Go escalates to SIGKILL once this elapses.
+	// on it forever.
+	cancelProcessTree(cmd)
+	// Backstop for anything the cancellation does not reach: abandon its output
+	// rather than hang the installer. Go escalates to SIGKILL once this elapses
+	// and closes the pipes, so Wait always returns.
 	cmd.WaitDelay = commandCancelGrace
 	runErr := cmd.Run()
 	if flusher, ok := r.Output.(interface{ Flush() }); ok {
--- a/cmd/installer/internal/core/command.go
+++ b/cmd/installer/internal/core/command.go
@@ -7,7 +7,13 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
+	"time"
 )
+
+// commandCancelGrace bounds how long a cancelled command may keep the
+// installer waiting on its output before it is abandoned.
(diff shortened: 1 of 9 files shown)


eetoc:
Confirmed and fixed in 890392c.

The report was accurate. I initially measured `GOOS=windows go build ./...` as passing and nearly dismissed the finding — that measurement was wrong: `go` on my machine is a gvm shell function that swallows the env prefix, so it silently built for darwin and exited 0. `GOOS=windows go env GOOS` returning `darwin` gave it away. Invoking the toolchain binary directly reproduces exactly what was predicted:

```
command.go:37:41: unknown field Setpgid in struct literal of type syscall.SysProcAttr
command.go:38:45: undefined: syscall.Kill
```

`main` builds for Windows, so this was a real regression in build surface even though the installer only ships linux binaries.

Fix follows the suggested shape: the process-group setup moved behind a `cancelProcessTree` helper with `//go:build unix` and `//go:build !unix` implementations. The non-unix path keeps the stdlib behaviour of signalling only the started process, and `WaitDelay` still bounds the wait — so the degradation there is a slower cancel, not a hang. `command_cancel_test.go` is unix-tagged as well since it drives `sh -c`.

Verified with `env GOOS=<os> <toolchain> build ./...` for windows, linux and darwin: all pass. Full suite still green; the cancellation test still completes in 0.2s.

---

One follow-up: the summary mentions "一处可改进的竞态条件" but no inline comment covers it, so I could not act on it. If it refers to reading `cmd.Process.Pid` inside the `Cancel` closure — `Cancel` is only invoked after a successful `Start`, so `Process` is non-nil by then — could you confirm, or point at the specific lines if it is something else?

innomentats (approved):
LGTM

PR #353, opened 2026-07-21
feat(installer): install the web UI on request, pre-pull the guest image, merge pull progress

eetoc:
> **Stacked on #352.** Base is `fix/installer-compose-progress-and-cancel`, so the diff here contains only the three feature commits. Please review #352 first; GitHub will retarget this to `main` once it merges.

Three related gaps, one per commit, each independently buildable and testable.

## 1. The web UI was unreachable through the installer

The frontend has been behind the `with-ui` Compose profile since 8c09a1a, but the installer never passed `--profile` and `filteredComposeEnvironment` strips `COMPOSE_PROFILES`. So the frontend could not be started through the installer at all — while the TUI still asked for a **Web UI port** and the summary still printed `http://localhost:<port>`, an address nothing was listening on.

- Install form gains an **install web UI** toggle, defaulting off to match the CLI (`--with-ui`).
- The port field stays visible but greyed out and skipped by the focus cycle while the UI is off, so the dependency is visible without the form changing height.
- The choice persists as `COMPOSE_PROFILES` in `.env`, mirroring how `COMPOSE_FILE` is already handled: an existing value wins unless the operator sets the flag explicitly, so **an upgrade cannot silently drop a frontend that is already deployed**. A later plain `docker compose up -d` also keeps the frontend.
- A URL is reported only when something serves it, and its host comes from the routing table rather than `localhost` — the installer usually runs on a server the operator reached over the network. Port 80 is left implicit. Without the UI the summary prints the command to enable it, and an explicit `--port` that cannot take effect is called out instead of silently ignored.

Text inputs alone could not express a toggle or a disabled field, so the form moved from `[]textinput.Model` to a small `formField` abstraction in its own file. `model.go` gets ~150 lines shorter.

## 2. The guest image was never pre-pulled

`DEFAULT_IMAGE` is a daemon setting, not a Compose service image — it is never interpolated into `docker-compose.yml`, so `compose pull` has no idea it exists. The guest image was fetched lazily on the first sandbox: install reported success in seconds, then the first agent run stalled on a several-hundred-megabyte download. When the registry is unreachable, the failure surfaced days after the install that caused it.

It is now pulled during install, with a form toggle and `--skip-guest-pull` to defer it.

The pull deliberately does **not** join the install transaction: the deployment is healthy without the guest image, so a failed download is a warning and the daemon still starts. Rolling back a working installation over a deferred convenience would cost more than it saves.

## 3. Pull progress flooded the TUI log

Docker repaints every active layer on each tick. `commandOutputWriter` maps `\r` to `\n` and strips the cursor sequences that would have overwritten the previous paint, so each repaint became a fresh line and one pull filled the entire 50-line log, pushing the actual installation steps out of view:

```
9319a554cac0 Downloading [==============>   ]  9.437MB/32.5MB
c3f793399f14 Downloading [=================>]  39.85MB/48.87MB
9319a554cac0 Downloading [==============>   ]  9.437MB/32.5MB
... the same two lines, repeatedly
```

Log entries are now keyed. A line carrying a layer id overwrites the entry with that key, so an in-flight layer occupies exactly one line updating in place; reaching a terminal status clears the key, freezing the line as history so completed layers accumulate above the active ones. Lines without a layer id are never merged.

Rendered check — the seven lines above collapse to three:

```
  Pulling images
  9319a554cac0 Pull complete
  c3f793399f14 Downloading [=============>     ]  44.1MB/48.87MB
```

## Verification

- Full suite green; `go vet` and `gofmt` clean. Each of the three commits builds and tests on its own, so `git bisect` stays usable.
- New coverage: profile persistence (fresh install on/off, upgrade preserving vs. explicit opt-out), guest pull and its skip, guest-pull failure not aborting the install, port field disabled/focus-skipped/re-enabled, progress merging, terminal-status freezing, non-layer lines never merged.
- Real run against Docker: install to `--no-start` produces the expected `.env` and summary.

> Reviewers on macOS: `internal/core` integration tests fail under the default `TMPDIR` because `/var` is a symlink and `validateInstallPath` rejects symlinked components. Pre-existing, reproduces on an unmodified tree — run with `TMPDIR` set to a non-symlinked path. Linux CI is unaffected.

## Deliberately out of scope

LLM configuration is **not** exposed in the form. The env var naming there needs sorting out first (`LLM_API_ENDPOINT` has no `OPENAI_BASE_URL` fallback while `LLM_API_KEY` does fall back to `OPENAI_API_KEY`, so an operator following OpenAI conventions gets a half-working config). Freezing one of those names into the installer form before that is settled would make it harder to change. `.env` stays the single source of truth for it.


eetoc:
Both findings confirmed and fixed in 763702c.

## `WithUI()` vs. profile lists

Real bug, and worse than the report suggests because `selectComposeProfiles` preserves an existing value verbatim — exactly the path where a hand-added profile appears. Probed before fixing:

```
"with-ui"           -> true
"with-ui,debug"     -> false   ← wrong
"debug,with-ui"     -> false   ← wrong
" with-ui , debug " -> false   ← wrong
```

`WithUI()` now scans the comma-separated list, trimming each entry. `with-ui-extra` still returns false, so it is not a substring match. The comment on `uiProfile` now states that the value is a list and that existing entries are preserved.

## Unconditional `PortSet` in the TUI

Also real: every TUI install with the UI toggled off raised the warning, and its "pass `--with-ui`" advice is meaningless there.

I took a different fix from the one suggested. Comparing the field against `core.DefaultOptions().Port` would treat "operator deliberately typed 80" and "operator never touched it" as the same thing, and it would still validate a field that cannot be edited. Instead the disabled port is skipped entirely in `readFields` — not parsed, not recorded. When the field is greyed out the operator has no way to put a choice in it, so there is nothing to interpret.

That also fixes a smaller problem the report did not mention: a malformed value left in the box before toggling the UI off used to block submission even though the field was inert.

## Tests

Four new tests, each verified to fail before the fix and pass after:

- `TestResultWithUIAcceptsProfileLists` — the table above, plus `""`, `"debug"`, `"with-ui-extra"`.
- `TestUpgradeKeepsWebUIAlongsideOtherProfiles` — installs with the UI, rewrites `.env` to `with-ui,debug`, upgrades without `WithUISet`, and asserts the profile survives, `WithUI()` holds, and a URL is still reported.
- `TestModelDisabledPortIsNotTreatedAsAChoice` — a garbage value in the disabled field neither blocks submission nor sets `PortSet`.
- `TestModelEnabledPortIsValidated` — the same field still validates and records once the UI is on.

Full suite green, `go vet` and `gofmt` clean, and cross-compilation still passes for windows, linux and darwin.

innomentats (approved):
LGTM

PR #364, opened 2026-07-22
fix(microsandbox): stop waiting on an exit that will never arrive

eetoc:
## Summary

- probe guest process liveness after two minutes of stream silence, and end the exec only when the process is definitely gone
- report a stream that ended without an exit status as a failure instead of exit code 0

## Why

A Microsandbox exec can end without ever reporting that it ended. The guest agent sends `ExecExited` only after both output pipes reach EOF (`crates/agentd/lib/session.rs`, pipe reader task: `while !stdout_eof || !stderr_eof { … }` and only then `child.wait()`), and a pipe reaches EOF only once *every* process that inherited it has exited. A command that leaves a background helper running therefore finishes while its stream stays open forever.

The driver had two guards and neither covered this:

- the guest-side `WithExecTimeout` is derived from the Go context deadline, which for a loader command is the whole run budget (`LOADER_RUN_TIMEOUT`, 8h in the affected deployment)
- the host-side drain window (`microsandboxExecExitIdleGracePeriod`) only arms **after** an exit event is seen, so the one path that never sees an exit is the one path with no deadline at all

Observed in production on a devboard task: the agent finished and published its MR at 09:55:49Z, the guest exited 0, and the run still sat in `running` two hours later holding 8 vCPU / 16 GB. Nothing was wrong with the task; a `ai-trs` proxy the entrypoint had started in the background was still holding the exec's stdout and stderr. The docker driver is unaffected because its exit code comes from the daemon API rather than from the stream, which is why the same loader script had worked for months.

## Approach

Silence alone proves nothing — a build can print nothing for an hour — so silence only triggers a question. After `microsandboxExecSilenceProbeInterval` with no events, ask the guest whether the pid from the `Started` event still exists, over a **separate** exec session that the stuck pipes cannot block. A live process means keep waiting. A probe that cannot answer also means keep waiting: "could not ask" and "it is gone" must not lead to the same decision. Only a definite answer ends the stream, as a failure that names the cause.

The second change is smaller and independent: `!sawExit` used to fall through to `exitCode = 0`, turning a lost status into a success whose output is quietly short. That is what surfaced downstream as `decode command result: no result payload found` — the result payload is printed last, so a truncated stream loses exactly it.

This is the runtime half of the problem. The loader-side half (the entrypoint leaking inherited descriptors) is fixed separately in agentkit; the two are independent, and this change also protects sandboxes agent-compose does not control the entrypoint of.

## Testing

- `go build`, `go vet`, `gofmt -l`, and `go test ./pkg/cache ./pkg/agentcompose/...` on darwin
- `go test -tags microsandboxcgo ./pkg/driver -run TestMicrosandbox -count=1` on linux/amd64 with cgo, all green
- `go test -tags microsandboxcgo ./pkg/driver -run TestMicrosandboxExecStream -count=1 -v` to confirm each new case runs rather than skips:
  - `TestMicrosandboxExecStreamStopsWhenProcessIsGoneWithoutExit`
  - `TestMicrosandboxExecStreamKeepsWaitingWhileProcessLives`
  - `TestMicrosandboxExecStreamKeepsWaitingWhenProbeFails`
  - `TestMicrosandboxExecStreamRejectsDoneWithoutExit`

## Checklist

### DIFF
--- a/pkg/driver/microsandbox_runtime.go
+++ b/pkg/driver/microsandbox_runtime.go
@@ -47,6 +47,22 @@ type microsandboxExecCollector struct {
 
 const microsandboxExecExitIdleGracePeriod = 2 * time.Second
 
+// microsandboxExecSilenceProbeInterval is how long a stream may produce nothing
+// at all before the guest is asked whether the process is still there.
+//
+// The guest agent reports a process exit only after both of its output pipes
+// reach EOF, and a pipe reaches EOF only once every process that inherited it
+// has exited. A command that leaves a background helper running therefore ends
+// without the stream ever saying so, and waiting on the next event is a wait
+// that never returns. Silence alone proves nothing -- a build can be quiet for
+// a long time -- so it only triggers the question, and the answer decides.
+const microsandboxExecSilenceProbeInterval = 2 * time.Minute
+
+// microsandboxExecLivenessProbeTimeout bounds the liveness question itself, so
+// a guest agent that has stopped answering cannot turn the probe into a second
+// place to hang.
+const microsandboxExecLivenessProbeTimeout = 30 * time.Second
+
 const (
 	microsandboxManagedLabel   = "agent-compose.managed"
 	microsandboxSandboxIDLabel = "agent-compose.sandbox_id"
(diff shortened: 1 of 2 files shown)


innomentats (approved):
LGTM

PR #369, opened 2026-07-22
fix(loaders): invalidate stale sticky sandboxes

winterfx:
## Summary

- fingerprint loader and effective sandbox configuration before reusing sticky bindings
- retire and replace sticky sandboxes when loader, workspace, agent, image, volume, or managed-project configuration changes
- persist the fingerprint with compare-and-swap binding updates to prevent concurrent replacement races
- add a versioned SQLite migration that preserves existing bindings and treats legacy empty hashes as stale once
- cover direct-loader and managed-project reuse, replacement, migration, and hash propagation paths

## Testing

- `go test ./pkg/storage/sqlite ./pkg/storage/configstore ./pkg/loaders ./pkg/runs ./pkg/agentcompose/adapters`
- `go test ./...` *(existing `pkg/volumes` macOS `/private/var` vs `/var` path assertions fail identically on `main` at `49a4578`)*


winterfx:
@monkeyscan 帮我review下这个分支的改动

eetoc (approved):
LGTM

winterfx:
https://github.com/chaitin/agent-compose/issues/104

PR #382, opened 2026-07-23
fix(microsandbox): recover exec streams after lost exits

eetoc:
## Summary

- keep exactly one Microsandbox exec receiver active while silence probes run, matching the SDK cancellation and concurrency contract
- stop and drain the receiver before closing an exec handle on terminal paths
- classify zombie guest processes as exited while preserving conservative behavior when process state cannot be read
- add regression coverage for receiver lifecycle, zombie detection, and the real lost-exit sequence

The failure required two conditions. A silence timeout previously canceled `Recv`, although the SDK documents that the underlying receive may continue in the background; a later `Recv` could then race for the exit event. Separately, `kill -0` reports a zombie PID as present, so a process that had exited but had not been reaped was treated as alive indefinitely.

## Testing

- `go test ./pkg/driver -count=1`
- `go test -race ./pkg/driver -count=1`
- `go vet ./pkg/driver`
- `task lint`
- `task build`
- `task test`
- Linux/amd64 race run with the `microsandboxcgo` build tag covering the exec-stream and zombie-probe tests
- Real KVM Microsandbox smoke regression:
  - kept the foreground command silent for 125 seconds so the first two-minute probe observed a live PID
  - then exited the foreground shell while a background process retained its output pipes
  - verified the second probe classified the unreaped zombie as exited and closed the stream with a bounded lost-exit error
  - completed successfully in 392.91 seconds including first-time base-disk preparation; the exec recovery itself completed after the expected two probe windows

## Checklist

- [x] Documentation is not required because no public API or configuration changed.
- [x] Tests added for receiver lifecycle, zombie detection, and real KVM behavior.
- [x] No secrets, private endpoints, certificates, or local runtime state included.

### DIFF
--- a/pkg/driver/microsandbox_runtime.go
+++ b/pkg/driver/microsandbox_runtime.go
@@ -71,6 +71,11 @@ const (
 type microsandboxExecEventReceiver func(context.Context) (*microsandbox.ExecEvent, error)
 type microsandboxExecHandleCloser func() error
 
+type microsandboxExecReceiveResult struct {
+	event *microsandbox.ExecEvent
+	err   error
+}
+
 // microsandboxExecLivenessProbe reports whether the guest process is still
 // present. A nil probe, or a probe that cannot answer, leaves the stream
 // waiting: only a definite "the process is gone" is allowed to end it.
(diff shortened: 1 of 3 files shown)


fuchencong (approved):
LGTM

PR #419, opened 2026-07-23
fix(llms): preserve Responses assistant output text

### LINKED ISSUE #415: [Bug]: LLM facade rewrites Responses assistant output_text to invalid input_text
## Summary

The LLM facade rewrites valid OpenAI Responses assistant-history content from
`output_text` to `input_text`. If an agent emits an assistant preamble before an MCP
tool call, the follow-up request that carries the tool result is rejected by the
upstream Responses API and the otherwise successful run ends with:

```text
Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.
```

This was found while retesting #323. It is a separate problem: the MCP server is
initialized, `tools/list` succeeds, and `tools/call` returns its result. The failure
happens only on the final model request after the tool call.

## Steps to reproduce

Use a daemon built from current `main`, configured with a Responses-compatible
upstream (`LLM_API_PROTOCOL=responses`), the Docker driver, and a Docker-reachable
`AGENT_COMPOSE_RUNTIME_BASE_URL`.

Create this minimal project:

```text
mcp-repro/
├── agent-compose.yml
└── mcp/
    └── stdio_server.py
```

`agent-compose.yml`:

```yaml
name: mcp-repro

workspaces:
  source:
    provider: file
    path: .

mcp_servers:
  example-search:
    type: local
    command: python3
    args:
      - /workspace/mcp/stdio_server.py

agents:
  repro:
    provider: codex
    image: ghcr.io/chaitin/agent-compose-guest:v2607.9.0
    driver:
      docker: {}
    workspace:
      name: source
    mcp_servers:
      - example-search
```

`mcp/stdio_server.py`:

```python
#!/usr/bin/env python3
import json
from pathlib import Path
import sys

log = Path("/workspace/mcp/requests.jsonl")

def reply(message_id, result):
    print(json.dumps({"jsonrpc": "2.0", "id": message_id, "result": result}), flush=True)

for line in sys.stdin:
    message = json.loads(line)
    method = message.get("method")
    params = message.get("params") or {}
    with log.open("a", encoding="utf-8") as stream:
        stream.write(json.dumps({
            "method": method,
            "name": params.get("name"),
            "arguments": params.get("arguments"),
        }) + "\n")

    message_id = message.get("id")
    if message_id is None:
        continue
    if method == "initialize":
        reply(message_id, {
            "protocolVersion": "2024-11-05",
            "capabilities": {"tools": {}},
            "serverInfo": {"name": "example-search", "version": "1.0.0"},
        })
    elif method == "tools/list":
        reply(message_id, {"tools": [{
            "name": "search_demo",
            "description": "Return a deterministic demo result.",
            "inputSchema": {
                "type": "object",
                "properties": {"query": {"type": "string"}},
                "required": ["query"],
                "additionalProperties": False,
            },
        }]})
    elif method == "tools/call":
        reply(message_id, {
            "content": [{"type": "text", "text": "mcp-tool-ok"}],
            "isError": False,
        })
    elif method == "ping":
        reply(message_id, {})
```

Apply the project and run a prompt that allows a pre-tool assistant preamble:

```bash
agent-compose --file agent-compose.yml up

agent-compose --file agent-compose.yml --json run repro --keep-running \
  --prompt 'Call the example-search/search_demo MCP tool once and report its result.'
```

The trigger is an assistant message before the function call, for example:

```text
I'll call the demo search tool once and report its result.
[mcp:example-search/search_demo]
{"query":"demo"}
mcp-tool-ok
```

Whether a model chooses to emit that preamble is nondeterministic. Once it does, the
facade defect itself is deterministic. The equivalent request shape before facade
rewriting contains valid assistant history:

```json
{
  "input": [
    {
      "role": "assistant",
      "type": "message",
      "content": [
        {"type": "output_text", "text": "I'll call the demo tool."}
      ]
    }
  ]
}
```

`RewriteRuntimeRequestForUpstream(..., ProtocolOpenAIResponses)` changes that part to:

```json
{"role":"assistant","content":[{"type":"input_text","text":"I'll call the demo tool."}]}
```

That rewritten assistant content is invalid for the standard Responses API.

## Expected behavior

The facade preserves `output_text` for assistant-history messages. The follow-up
Responses request succeeds and the run returns the MCP result.

For standard Responses content:

- user/developer input text uses `input_text`;
- assistant history uses `output_text`;
- provider-specific generic `text` normalization remains limited to the provider mode
  that requires it.

## Actual behavior

The MCP lifecycle completes through `tools/call`, and the fixed tool result appears in
the transcript. The final request then contains assistant content typed as `input_text`.
The upstream rejects the initial request and Codex's stream retry with the same error.
The stream closes without `response.completed`, so the run exits with status 1:

```text
Error: stream disconnected before completion: stream closed before response.completed
```

Observed MCP methods before failure:

```json
{"method":"initialize"}
{"method":"notifications/initialized"}
{"method":"tools/list"}
{"method":"tools/call","name":"search_demo","arguments":{"query":"demo"}}
```

## Root cause

`normalizeRuntimeRawResponsesContent` in `pkg/llms/facade_bridge.go` rewrites every
`output_text` part to the selected default text type without checking the item role.
For standard Responses requests that default is `input_text`, so valid assistant
history becomes invalid input content.

The existing test
`TestRewriteRuntimeRequestForUpstreamNormalizesResumedAssistantText` currently asserts
the invalid result (`assistant/output_text` becomes `input_text`). The unconditional
rewrite was introduced by commit `49f6acfb` (`fix(llms): normalize OpenCode resumed
response text`) and is present on current `main`.

The normalization should be role-aware. A regression test should cover an assistant
preamble followed by a function call and tool output, and assert that the subsequent
standard Responses request keeps assistant text as `output_text`.

## Version or commit

- Reproduced on `main` commit `92b5bd307d3c10b72ae112f3ebd3f8628e908001`
- First known introducing commit: `49f6acfbdf713e7d62e6c60f846cb54810714838`
- Guest image used: `ghcr.io/chaitin/agent-compose-guest:v2607.9.0`
- Codex CLI in the guest: `0.144.1`
- Runtime driver: Docker

No private endpoint, credential, token, internal hostname, or application-specific data
is required for this reproduction.

fuchencong:
## Summary

- make OpenAI Responses request normalization role-aware so assistant history keeps `output_text` while user and developer input uses `input_text`
- preserve provider-specific generic `text` normalization only for providers that explicitly require it
- add regression coverage for an assistant preamble followed by a function call and tool output, plus standard and generic provider text-type normalization

Fixes #415.

## Testing

  - Unit: 74.53%
  - Integration: 61.02%
  - E2E: 61.45%
  - Combined: 78.94%
- `GOFLAGS=-buildvcs=false go test ./pkg/llms -count=1`
- `GOFLAGS=-buildvcs=false go test ./pkg/agentcompose/proxy -count=1`
- `data/ai_analyse/ai-test-suit/suit-mcp` case 14 (`real_codex_issue_323_exact.sh`)
  - run exited successfully
  - MCP `tools/call` and runtime MCP event observed
  - deterministic tool result returned
  - `--rm` removed the sandbox

## Checklist

- [x] Documentation updated when behavior or configuration changed (not required; this fixes protocol normalization without changing public configuration).

### DIFF
--- a/pkg/llms/facade_bridge.go
+++ b/pkg/llms/facade_bridge.go
@@ -53,13 +53,8 @@ func normalizeRuntimeRawResponsesInput(payload map[string]json.RawMessage, gener
 		return false
 	}
 	var changed bool
-	defaultTextType := "input_text"
-	if genericTextParts {
-		defaultTextType = "text"
-	}
-	textTypeJSON, _ := json.Marshal(defaultTextType)
 	for _, item := range items {
-		if normalizeRuntimeRawResponsesContent(item, textTypeJSON) {
+		if normalizeRuntimeRawResponsesContent(item, genericTextParts) {
 			changed = true
 		}
 	}
@@ -74,7 +69,7 @@ func normalizeRuntimeRawResponsesInput(payload map[string]json.RawMessage, gener
 	return true
 }
 
-func normalizeRuntimeRawResponsesContent(item map[string]json.RawMessage, textType []byte) bool {
+func normalizeRuntimeRawResponsesContent(item map[string]json.RawMessage, genericTextParts bool) bool {
 	raw := item["content"]
 	if len(raw) == 0 || string(raw) == "null" {
 		return false
@@ -83,18 +78,30 @@ func normalizeRuntimeRawResponsesContent(item map[string]json.RawMessage, textTy
 	if err := json.Unmarshal(raw, &parts); err != nil {
 		return false
 	}
+	textType := "input_text"
+	if genericTextParts {
+		textType = "text"
+	} else {
+		var role string
+		if err := json.Unmarshal(item["role"], &role); err == nil && role == string(protocolbridge.RoleAssistant) {
+			textType = "output_text"
+		}
+	}
+	textTypeJSON, _ := json.Marshal(textType)
 	var changed bool
 	for _, part := range parts {
 		if len(part["text"]) == 0 || string(part["text"]) == "null" {
 			continue
 		}
 		if len(part["type"]) == 0 || string(part["type"]) == "null" {
-			part["type"] = textType
+			part["type"] = textTypeJSON
 			changed = true
 			continue
 		}
-		if string(part["type"]) == `"output_text"` || (string(textType) == `"text"` && string(part["type"]) == `"input_text"`) {
-			part["type"] = textType
+		var partType string
+		if err := json.Unmarshal(part["type"], &partType); err == nil &&
+			(partType == "input_text" || partType == "output_text") && partType != textType {
+			part["type"] = textTypeJSON
 			changed = true
 		}
 	}
--- a/pkg/llms/facade_bridge_test.go
+++ b/pkg/llms/facade_bridge_test.go
@@ -7,13 +7,13 @@ import (
 	protocolbridge "github.com/chaitin/ai-api-protocol-bridge"
 )
 
-func TestRewriteRuntimeRequestForUpstreamNormalizesResumedAssistantText(t *testing.T) {
+func TestRewriteRuntimeRequestForUpstreamPreservesAssistantTextAfterToolCall(t *testing.T) {
 	body := []byte(`{
-		"model":"old",
 		"input":[
-			{"role":"user","content":[{"type":"input_text","text":"remember kiwi"}]},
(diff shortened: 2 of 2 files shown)


eetoc (approved):
LGTM

PR #428, opened 2026-07-24
fix(projects): preserve legacy loader environment

zhuanyongxigua:
## Summary

- preserve the task-local environment of adopted v1 Loaders when the synthetic `legacy-v1-default` project is migrated or reapplied
- reconcile Loader environment keys against the previous managed Agent environment so unchanged ProjectSpec keys retain out-of-band Loader edits, while explicit Agent env additions, updates, deletions, and `secret` metadata changes take effect
- retain the existing runtime precedence of global environment, Agent environment, Loader environment, and per-request environment
- add focused unit, SQLite integration, E2E-shape, and runtime-precedence regression coverage
- document the compatibility overlay and reconciliation rule

This fixes whole-project `ApplyProject` calls rebuilding an adopted Loader from `AgentSpec.env` and silently replacing Loader-only environment values. It does not change the public API, protobuf schema, or database schema.

## Testing

- `go test ./pkg/projects -count=1`
- `go test ./pkg/agentcompose/adapters -count=1`
- `go test ./cmd/... ./pkg/... ./proto/health/v1 ./proto/health/v1/healthv1connect ./proto/agentcompose/v2 ./proto/agentcompose/v2/agentcomposev2connect`
- `GOPROXY=off GOFLAGS=-mod=vendor go test ./pkg/projects -run '^TestIntegrationLegacyLoaderEnvironmentSurvivesProjectApply$' -count=1 -v` on an isolated Linux checkout
- `GOPROXY=off GOFLAGS=-mod=vendor go test ./pkg/agentcompose/adapters -run '^TestLoaderSandboxRunnerEnvironmentPrecedence$' -count=1 -v` on an isolated Linux checkout
- `task lint`
- `task build`
- `git diff --check`

The single local `task test` wrapper was not used because its deployment and
coverage helpers require GNU `sed` and a modern Bash, while the development
host provides BSD `sed` and Bash 3.2. The PR's Linux CI passed the coverage,
Go, lint, runtime, installer, binary/image build, and full-image Docker smoke
jobs.

## Checklist

- [x] Documentation updated when behavior or configuration changed.
- [x] Tests added or updated for user-visible behavior.
- [x] No secrets, private endpoints, internal certificates, or local runtime state included.

PR #430, opened 2026-07-24
fix(projects): bind adopted loaders to managed agents

zhuanyongxigua:
## Summary

- bind an adopted legacy Loader to the Project-managed Agent while preserving the legacy Loader ID, history, trigger state, workspace, and Loader-level environment overlay
- repair Loader records persisted by older versions on the next ordinary `ApplyProject`
- add an opt-in host-daemon E2E that applies Projects through the public API and runs real Docker Scheduler sandboxes
- document the managed Agent binding and the reusable E2E command

This is a follow-up to #428. The environment overlay was persisted correctly,
but a real Scheduler run showed that the adopted Loader still referenced the
original v1 Agent. As a result, later ProjectSpec Agent environment changes
were stored on the managed Agent but were not used by the sandbox runtime.

The production behavior change is intentionally small: the override merge now
keeps the Agent binding compiled from the current ProjectSpec instead of
restoring the legacy Agent ID.

## Side-effect coverage

The host-daemon E2E uses an isolated database and daemon, public Connect APIs,
and two short-lived Docker Scheduler sandboxes. It verifies:

- dry-run does not mutate the Project revision or Loader environment
- script-only whole-Project apply preserves out-of-band legacy Loader edits
- old persisted v1 Agent bindings are repaired on apply
- explicit Agent env add/change/delete/secret changes take effect
- effective precedence remains global < Agent < Loader < per-request
- Loader-only keys and unchanged Loader overrides remain effective
- an ordinary v2 Project remains ProjectSpec-authoritative and isolated from
  the synthetic legacy Project
- sandboxes are removed through the public API and leave no Docker residue

Focused unit and integration coverage also verifies that Loader identity,
history, trigger timestamps/state, workspace, and the managed Agent binding are
preserved together.

## Testing

- `go test ./pkg/projects -count=1`
- `go test ./pkg/agentcompose/adapters -count=1`
- `go test ./test/e2e -run '^TestE2ELegacyLoaderEnvironmentSurvivesPublicProjectApply$' -count=1`
- `AGENT_COMPOSE_E2E_LEGACY_LOADER_ENV_IMAGE=<existing-image> go test ./test/e2e -run '^TestE2ELegacyLoaderEnvironmentSurvivesPublicProjectApply$' -count=1 -v` on an isolated Linux Docker host
- `go test ./pkg/projects -count=1` on an isolated Linux host
- `go test ./pkg/agentcompose/adapters -run '^TestLoaderSandboxRunnerEnvironmentPrecedence$' -count=1` on an isolated Linux host
- `task lint`
- `task build:agent-compose`
- `task build:proto`
- `git diff --check`

## Checklist

- [x] Documentation updated for the behavior and opt-in E2E.
- [x] Regression tests cover persistence, public API, runtime precedence, and cleanup.
- [x] No API, protobuf, or database schema change.
- [x] No environment-specific hostnames, paths, credentials, or runtime state included.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -121,6 +121,24 @@ compiled only with the `docker_e2e` build tag, so the ordinary `task test`
 coverage gate does not include this scheduler Docker E2E or create its runtime
 containers.
 
+The legacy Loader environment migration has an opt-in host-daemon E2E because
+real `ApplyProject` validates Docker image availability. It seeds an isolated
+legacy database, starts the candidate daemon, and exercises dry-run and real
+updates through the public ProjectService while checking both legacy and
+ordinary v2 Project persistence:
+
+```bash
+AGENT_COMPOSE_E2E_LEGACY_LOADER_ENV_IMAGE=agent-compose-guest:latest \
+  go test ./test/e2e \
+    -run '^TestE2ELegacyLoaderEnvironmentSurvivesPublicProjectApply$' \
+    -count=1 -v
+```
+
+The selected image must already exist in the local Docker daemon. The test
+starts two short-lived Scheduler sandboxes to verify the effective
+global/Agent/legacy Loader/per-call environment precedence, then removes them
+through the public API and checks Docker cleanup.
+
 The full daemon image Docker lifecycle E2E is opt-in because it starts the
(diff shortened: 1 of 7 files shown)


eetoc (approved):
LGTM

PR #443, opened 2026-07-27
fix(storage): make webhook event commits unambiguous

### LINKED ISSUE #439: Webhook ingestion can report failure after the event insert commits
### Summary

Webhook ingestion can return HTTP 500 even though the event row was committed successfully. A caller that retries the apparently failed request can therefore create and dispatch a duplicate event when it did not provide an idempotency key.

The ambiguity comes from `CreateEvent`: it performs the INSERT, reads `LastInsertId`, and then calls `GetEvent` with the original request context. Under database contention, the INSERT may complete near the request deadline while the follow-up SELECT fails because the context has expired. The handler then reports `failed to store webhook event`, although the dispatcher can subsequently observe and publish the committed row.

The shared SQLite database is also configured with one open connection, so an unrelated slow database operation can cause head-of-line blocking across webhook ingestion, run persistence, and sandbox projection work.

### Steps to reproduce

1. Run the daemon with the SQLite store.
2. Occupy or delay the sole database connection until a webhook request is close to its context deadline.
3. POST a valid webhook without an idempotency key.
4. Allow the INSERT to complete, but let the follow-up `GetEvent` observe the expired context.
5. Observe an HTTP 500 response while the committed event is later dispatched.
6. Retry the webhook and observe that a second event may be created.

### Expected behavior

Once the event INSERT commits, webhook ingestion should return an accepted response for that event and must not report that storage failed because of a redundant post-insert read. Storage behavior should avoid ambiguous outcomes that cause callers to retry committed work.

Consider returning the normalized inserted record after `LastInsertId`, or using `INSERT ... RETURNING`, plus regression coverage for context cancellation immediately after commit. Separately, consider making SQLite connection concurrency configurable and adding connection-wait/slow-query observability.

### Actual behavior

The INSERT can commit, the post-insert SELECT can fail with the expired request context, and the endpoint returns HTTP 500. The dispatcher still publishes the committed event, so a caller retry can produce duplicate execution.

### Version or commit

Current main

### Runtime driver

not runtime-related

### Logs

```text
POST /api/webhooks/<topic> status=500 latency≈request deadline
error: failed to store webhook event

The inserted event is subsequently visible with a published dispatch status.
```

winterfx:
## Summary

- make the file-backed SQLite runtime pool configurable through `SQLITE_MAX_OPEN_CONNS`, defaulting to 4 while keeping migrations and in-memory databases on one connection
- finalize event sequence, payload, and payload hash in one transaction so dispatchers cannot observe a pending event with `sequence: 0`
- return the committed event directly and remove the webhook handler's post-commit read/write path, avoiding HTTP 500 after a successful commit
- preserve large JSON integers and idempotent retries while finalizing payload sequences

Fixes #439

## Testing

- `go test ./...` (run against the exact staged snapshot in a temporary worktree)
- `go test -count=2 -race ./pkg/storage/configstore ./pkg/events/webhooks ./pkg/loaders`
- `task lint`
- `task build`
- `task docs:build`

`task test` was not run because this macOS host does not provide GNU `sha256sum`; the complete Go suite was run directly as listed above.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,9 @@ HTTP_LISTEN=0.0.0.0:7410
 AGENT_COMPOSE_AUTH_TOKEN=
 DATA_ROOT=/data
 # SANDBOX_ROOT=/data/sandboxes
+# Maximum runtime connections to the shared SQLite database. Startup migrations
+# always use one connection, and in-memory databases always remain single-connection.
+# SQLITE_MAX_OPEN_CONNS=4
 
 # Jupyter proxy base path. Keep this in sync with any external reverse proxy.
 JUPYTER_PROXY_BASE=/jupyter
--- a/cmd/agent-compose/cli_daemon_lifecycle_test.go
+++ b/cmd/agent-compose/cli_daemon_lifecycle_test.go
@@ -281,7 +281,7 @@ agents:
 		t.Fatalf("create test data root: %v", err)
 	}
 	storeConfig := &config.Config{DataRoot: dataRoot, DbAddr: filepath.Join(dataRoot, "data.db")}
-	database, err := storagesqlite.Open(storeConfig.DbAddr, storeConfig.DbTimeout)
+	database, err := storagesqlite.OpenWithMaxOpenConns(storeConfig.DbAddr, storeConfig.DbTimeout, storeConfig.EffectiveSQLiteMaxOpenConns())
 	if err != nil {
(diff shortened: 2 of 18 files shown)


winterfx:
@monkeyscan 帮我review

winterfx:
@monkeyscan 帮我review

eetoc (approved):
LGTM

PR #453, opened 2026-07-27
fix(cli): reject conflicting run cleanup flags

### LINKED ISSUE #450: [Bug]: run silently ignores --rm when combined with --keep-running
### Summary

`agent-compose run` accepts `--rm` and `--keep-running` at the same time, even though the flags request mutually exclusive sandbox lifecycle behavior.

The command succeeds and silently gives `--keep-running` precedence. As a result, `--rm` is ignored and the sandbox remains running after completion.

### Steps to reproduce

1. Save the following project as `conflicting-cleanup.yml`:

```yaml
name: conflicting-run-cleanup-repro

agents:
  runner:
    provider: codex
```

2. Apply it with an agent-compose daemon running and a default guest image configured:

```bash
agent-compose -f conflicting-cleanup.yml up
```

3. Pass both cleanup flags to one command run:

```bash
agent-compose \
  -f conflicting-cleanup.yml \
  --json \
  run runner \
  --rm \
  --keep-running \
  --command 'printf "conflicting-cleanup-flags\n"'
```

### Expected behavior

The CLI should reject the mutually exclusive flags before creating a run or sandbox, return a non-zero usage exit code, and print a clear diagnostic such as:

```text
run --rm cannot be combined with --keep-running
```

The result should not depend on flag order or an undocumented implementation priority.

### Actual behavior

The command returns exit code 0 and reports a successful run:

```json
{
  "status": "succeeded",
  "exit_code": 0,
  "output": "conflicting-cleanup-flags\n",
  "sandbox_id": "<new-sandbox-id>"
}
```

The new sandbox remains in the running state. `--keep-running` wins and `--rm` has no effect.

The regression probe produced:

```text
{
  "id": "e3db62ec916419ca7253812ea9816ef66bb28af58e42340115652d8488256a10",
  "status": "succeeded",
  "sandbox_id": "ef47401a4151ac4c33b5292f98a62959361ccf7db7bd0636a0228e42179c496a",
  "exit_code": 0,
  "output": "conflicting-cleanup-flags-unexpected\n"
}
run-exit-code=0
```

The probe sandbox was explicitly removed after the test.

### Impact

- The CLI accepts a contradictory request instead of reporting user error.
- Automation may expect `--rm` to remove a sandbox, while the sandbox remains running.
- Retained runtimes can consume resources and preserve files or sensitive state longer than intended.
- Behavior is controlled by implementation priority rather than an explicit CLI contract.

### Version or commit

- Platform: Linux amd64

### Runtime driver

Reproduced with Docker. The defect is in generic CLI option validation and cleanup-policy selection, so it is not believed to be driver-specific.

### Additional analysis

Both flags are independent booleans in `composeRunOptions` and are registered without a mutual-exclusion rule.

`normalizeComposeRunOptions` currently validates only the `--driver`/`--sandbox` conflict and returns successfully when both cleanup flags are true.

Cleanup policy selection then uses this fixed priority:

```go
cleanupPolicy := agentcomposev2.RunSandboxCleanupPolicy_RUN_SANDBOX_CLEANUP_POLICY_STOP_ON_COMPLETION
if normalizedOptions.KeepRunning {
    cleanupPolicy = agentcomposev2.RunSandboxCleanupPolicy_RUN_SANDBOX_CLEANUP_POLICY_KEEP_RUNNING
} else if normalizedOptions.Remove {
    cleanupPolicy = agentcomposev2.RunSandboxCleanupPolicy_RUN_SANDBOX_CLEANUP_POLICY_REMOVE_ON_COMPLETION
}
```

Because `KeepRunning` is checked first, `--rm` is silently ignored whenever both booleans are true. This is independent of the order in which the flags appear on the command line.

The smallest fix is to reject the combination in `normalizeComposeRunOptions` (or through Cobra's mutually-exclusive flag support) and add a CLI test confirming that validation happens before any run/sandbox request is sent.

Relevant implementation paths:

- `cmd/agent-compose/cli_run_definition.go`: registers `--keep-running` and `--rm` as separate boolean flags.
- `cmd/agent-compose/cli_run_command.go`: `normalizeComposeRunOptions` does not reject the combination; cleanup policy selection prioritizes `KeepRunning`.

### Regression evidence

- Case: `20a-conflicting-run-cleanup-flags`
- Run ID: `e3db62ec916419ca7253812ea9816ef66bb28af58e42340115652d8488256a10`
- Sandbox ID: `ef47401a4151ac4c33b5292f98a62959361ccf7db7bd0636a0228e42179c496a`

fuchencong:
## Summary

- reject `agent-compose run --rm --keep-running` as an invalid CLI request before creating service clients or sending a run request
- return a clear usage error instead of silently giving `--keep-running` precedence
- cover both flag orders with regression tests

Fixes #450

## Testing

- `GOFLAGS=-buildvcs=false go test ./cmd/agent-compose -run '^TestCLIRunInputModeUsageErrors$' -count=1`
- `GOFLAGS=-buildvcs=false go test ./cmd/agent-compose -count=1`
  - Unit: 75.02%
  - Integration: 60.80%
  - E2E: 60.59%
  - Combined: 79.12%

## Checklist

- [x] Documentation reviewed; no update is required for validation of an invalid flag combination.

### DIFF
--- a/cmd/agent-compose/cli_run_command.go
+++ b/cmd/agent-compose/cli_run_command.go
@@ -432,6 +432,9 @@ func (w *terminalStreamWriter) Finish() error {
 func normalizeComposeRunOptions(cmd *cobra.Command, options composeRunOptions) (composeRunOptions, error) {
 	options.SandboxID = strings.TrimSpace(options.SandboxID)
 	options.Driver = strings.TrimSpace(options.Driver)
+	if options.Remove && options.KeepRunning {
+		return options, commandExitError{Code: exitCodeUsage, Err: fmt.Errorf("run --rm cannot be combined with --keep-running")}
+	}
 	if options.Driver != "" {
 		driver, err := driverpkg.ResolveSandboxRuntimeDriver(options.Driver, "")
 		if err != nil {
--- a/cmd/agent-compose/cli_run_command_test.go
+++ b/cmd/agent-compose/cli_run_command_test.go
@@ -105,6 +105,16 @@ agents:
 			want: "run --driver cannot be combined with --sandbox",
 		},
 		{
+			name: "remove with keep running",
+			args: []string{"run", "--host", server.URL, "--file", composePath, "reviewer", "--rm", "--keep-running", "--prompt", "check"},
+			want: "run --rm cannot be combined with --keep-running",
+		},
+		{
+			name: "keep running with remove",
+			args: []string{"run", "--host", server.URL, "--file", composePath, "reviewer", "--keep-running", "--rm", "--prompt", "check"},
+			want: "run --rm cannot be combined with --keep-running",
+		},
+		{
 			name: "command and prompt flags",
 			args: []string{"run", "--host", server.URL, "--file", composePath, "reviewer", "--command", "echo hi", "--prompt", "check"},
 			want: "only one of --prompt or --command",


eetoc (approved):
LGTM

PR #460, opened 2026-07-27
fix(cli): reject conflicting exec targets

### LINKED ISSUE #457: [Bug]: exec ignores positional sandbox when deprecated --run is also provided
### Summary

`agent-compose exec <sandbox-a> --run <run-from-sandbox-b>` accepts two conflicting execution targets. The positional sandbox is silently ignored, the deprecated `--run` target wins, and the command executes in run B's sandbox.

This is a high-risk target-selection bug: the command line visibly names sandbox A, but filesystem reads, writes, secret access, and destructive commands can occur in sandbox B.

### Minimal configuration

Save the following as `exec-selector-conflict.yml`:

```yaml
name: exec-selector-conflict-repro

agents:
  runner:
    provider: codex
```

### Steps to reproduce

1. With an agent-compose daemon and a default guest image configured, apply the project:

```bash
agent-compose -f exec-selector-conflict.yml up
```

2. Create two separate running sandboxes with distinct marker files:

```bash
run_a_json=$(
  agent-compose -f exec-selector-conflict.yml --json \
    run runner --keep-running \
    --command 'printf "sandbox-a\n" > /tmp/exec-target-marker'
)

run_b_json=$(
  agent-compose -f exec-selector-conflict.yml --json \
    run runner --keep-running \
    --command 'printf "sandbox-b\n" > /tmp/exec-target-marker'
)

sandbox_a=$(printf '%s\n' "$run_a_json" | jq -r '.sandbox_id')
sandbox_b=$(printf '%s\n' "$run_b_json" | jq -r '.sandbox_id')
run_b=$(printf '%s\n' "$run_b_json" | jq -r '.id')

printf 'sandbox A: %s\nsandbox B: %s\nrun B: %s\n' \
  "$sandbox_a" "$sandbox_b" "$run_b"
test "$sandbox_a" != "$sandbox_b"
```

3. Name sandbox A positionally while also passing a run that belongs to sandbox B:

```bash
agent-compose \
  -f exec-selector-conflict.yml \
  exec "$sandbox_a" \
  --run "$run_b" \
  --command 'cat /tmp/exec-target-marker'
printf 'exec exit code: %s\n' "$?"
```

Observed output:

```text
Warning: agent-compose exec --run is deprecated and will be removed in a future release; use agent-compose exec <sandbox> instead.
sandbox-b
exec exit code: 0
```

The `sandbox-b` marker proves that the positional sandbox A was ignored and the command ran in sandbox B.

4. Clean up the two probe sandboxes:

```bash
agent-compose -f exec-selector-conflict.yml rm --force "$sandbox_a"
agent-compose -f exec-selector-conflict.yml rm --force "$sandbox_b"
```

### Expected behavior

The CLI should reject positional `<sandbox>` together with deprecated `--run` before resolving either target or sending an Exec RPC. For example:

```text
exec target can only be specified once: use either <sandbox> or --run
```

No command should execute in either sandbox when target selection is ambiguous.

### Actual behavior

The command exits with code 0. It prints only a deprecation warning, ignores positional sandbox A, resolves run B, and executes the command in run B's sandbox.

### Impact

- A command can modify or delete files in the wrong sandbox.
- It can read secrets, credentials, mounts, or workspace data from a runtime the user did not intend to target.
- Automation that migrates from deprecated `--run` to positional sandbox syntax can accidentally retain both arguments and execute against the legacy target.
- The visible positional target gives operators false confidence about where a destructive command ran.

### Version/commit

- Platform: Linux amd64

### Runtime driver

Reproduced with Docker. The selector bug is in generic CLI argument validation and request normalization, so it is not believed to be driver-specific.

### Additional analysis

Two paths combine to produce the behavior:

1. `composeExecArgs` in `cmd/agent-compose/cli_exec_command.go` returns success immediately whenever `--run` was supplied, so it does not enforce the positional target shape:

   ```go
   if cmd.Flags().Changed("run") {
       return nil
   }
   return cobra.MinimumNArgs(1)(cmd, args)
   ```

2. `normalizeComposeExecRequest` enters the legacy-target branch when `--run` is present. That branch builds an `ExecRequest_RunId` from `options.RunID` and never checks or consumes the positional sandbox in `args`.

The existing `len(legacyTargetFlags) > 1` check cannot catch this combination because the positional sandbox is not included in `legacyTargetFlags`.

The fix should reject `--run` whenever positional arguments include a sandbox, before client creation/RPC. A regression test should use two distinct sandboxes and verify both the non-zero usage result and that the payload was not executed in either sandbox.

### Regression evidence

- Case: `41c-exec-target-conflict`
- Positional sandbox ID: `c1b623100445d64d6c1c11f47442c9dd266719425e93afbafa49e7453f49d4cd`
- Supplied run ID: `7c5aefad9ef594798f6355de47a29fbe4980d143a3b59f6d19f031ae128e0b57`
- That run's sandbox ID: `983a5e76088360e0d8a3b3b38776c39f02680e05ce918aac921f48acdca4754d`
- Observed result: exit code `0`; the command printed `exec-target-conflict-unexpected` after resolving the `--run` target.

fuchencong:
## Summary

- reject `agent-compose exec <sandbox> --run <run-id>` during Cobra argument validation, before project or target resolution and before any Exec RPC
- enforce the same target-selection invariant during request normalization so internal callers cannot silently prefer the deprecated run target
- add unit and in-process integration regressions that verify usage exit code 2, no deprecation warning, and zero `ExecStream` calls for conflicting targets
- document the mutual exclusion in the English and Chinese CLI manuals

Closes #457.

## Testing

- `GOFLAGS=-buildvcs=false go test ./cmd/agent-compose -run '^(TestComposeExecRejectsConflictingTargets|TestIntegrationCLIExecRejectsConflictingTargetsBeforeRPC)$' -count=1 -v`
- `GOFLAGS=-buildvcs=false go test ./cmd/agent-compose -count=1`
- `task docs:build`
- `GOFLAGS=-buildvcs=false task test GO_BUILD_CACHE_DIR="$GOCACHE"` (with inherited agent-compose product configuration cleared in the test subshell)

Coverage from the full test gate:

- Unit: 75.12%
- Integration: 60.74%
- E2E: 60.49%
- Combined: 79.12%

## Checklist

### DIFF
--- a/cmd/agent-compose/cli_exec_command.go
+++ b/cmd/agent-compose/cli_exec_command.go
@@ -77,10 +77,20 @@ func executeComposeRunRequest(cmd *cobra.Command, cli cliOptions, projectName, p
 }
 
 func composeExecArgs(cmd *cobra.Command, args []string) error {
+	if err := validateComposeExecTargetArgs(cmd, args); err != nil {
+		return err
+	}
 	if cmd.Flags().Changed("run") {
 		return nil
 	}
 	return cobra.MinimumNArgs(1)(cmd, args)
+}
+
+func validateComposeExecTargetArgs(cmd *cobra.Command, args []string) error {
+	if cmd.Flags().Changed("run") && len(args) > 0 {
+		return commandExitError{Code: exitCodeUsage, Err: fmt.Errorf("exec target can only be specified once: use either <sandbox> or --run")}
+	}
+	return nil
 }
 
 func runComposeExecCommand(cmd *cobra.Command, cli cliOptions, options composeExecOptions, args []string) error {
@@ -204,6 +214,9 @@ func (p promptAttachInputPrompt) String() string {
 }
 
 func normalizeComposeExecRequest(cmd *cobra.Command, clients cliServiceClients, projectID string, options composeExecOptions, args []string) (*agentcomposev2.ExecRequest, error) {
+	if err := validateComposeExecTargetArgs(cmd, args); err != nil {
+		return nil, err
+	}
 	commandText := strings.TrimSpace(options.Command)
 	promptText := strings.TrimSpace(options.Prompt)
 	if commandText != "" && promptText != "" {
--- a/cmd/agent-compose/cli_exec_command_test.go
+++ b/cmd/agent-compose/cli_exec_command_test.go
@@ -70,6 +70,28 @@ func TestComposeExecCommandFromPositionalArgs(t *testing.T) {
 	}
 }
 
+func TestComposeExecRejectsConflictingTargets(t *testing.T) {
+	cmd := &cobra.Command{Use: "exec"}
+	cmd.Flags().String("run", "", "")
+	if err := cmd.Flags().Set("run", "run-b"); err != nil {
+		t.Fatalf("set exec run: %v", err)
+	}
+
+	args := []string{"sandbox-a"}
+	if err := composeExecArgs(cmd, args); commandExitCode(err) != exitCodeUsage || !strings.Contains(err.Error(), "use either <sandbox> or --run") {
+		t.Fatalf("composeExecArgs conflict err=%v code=%d", err, commandExitCode(err))
+	}
+
+	var stderr bytes.Buffer
+	cmd.SetErr(&stderr)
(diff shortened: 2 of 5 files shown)


eetoc (approved):
LGTM

PR #461, opened 2026-07-27
fix(sandbox): reject unknown status filters

### LINKED ISSUE #455: [Bug]: ps silently accepts unknown sandbox status filters
### Summary

`agent-compose ps --status` accepts any non-empty status string. A typo or unknown value therefore exits successfully and returns an empty result instead of reporting invalid CLI input.

For example, `--status definitely-invalid` returns exit code 0 with `"sandboxes": null`. This is indistinguishable from a legitimate query that has no matching sandboxes and can make automation reach the wrong conclusion.

### Minimal configuration

Save the following as `invalid-ps-status.yml`:

```yaml
name: invalid-ps-status-repro

agents:
  runner:
    provider: codex
```

### Steps to reproduce

1. With an agent-compose daemon running, apply the project:

```bash
agent-compose -f invalid-ps-status.yml up
```

2. Query a status that is not a supported sandbox state and capture the exit code:

```bash
agent-compose \
  -f invalid-ps-status.yml \
  --json \
  ps --all --status definitely-invalid
printf 'ps exit code: %s\n' "$?"
```

Observed output:

```json
{
  "project": {
    "name": "invalid-ps-status-repro"
  },
  "sandboxes": null
}
```

```text
ps exit code: 0
```

The behavior does not depend on whether the project currently has sandboxes: the unsupported status is accepted as a normal filter value.

### Expected behavior

The CLI should reject unsupported values before making the list request, return a non-zero usage exit code, and identify the accepted sandbox statuses. For example:

```text
invalid --status "definitely-invalid": expected one or more supported sandbox statuses
```

Comma-separated status input should validate every non-empty item so that one typo cannot silently turn the query into an empty result.

### Actual behavior

The command exits with code 0, prints no diagnostic, and returns `"sandboxes": null`.

### Impact

- A misspelled status is hidden as a successful query with no matches.
- Cleanup and monitoring scripts may incorrectly conclude that no sandbox is running, stopped, or failed.
- CI checks can pass while querying a status that the product never produces.
- The behavior is inconsistent with CLI filters that validate their accepted values.

### Version/commit

- Platform: Linux amd64

### Runtime driver

Reproduced with Docker. The defect is in generic CLI status-filter validation and is not believed to be runtime-driver-specific.

### Additional analysis

`composePSStatusFilter` in `cmd/agent-compose/cli_project_display.go` lowercases and trims every comma-separated item, then adds every non-empty string to the filter map:

```go
for _, value := range values {
    value = strings.ToLower(strings.TrimSpace(value))
    if value == "" {
        continue
    }
    result[value] = true
}
```

There is no membership check against the sandbox statuses supported by the CLI/API. The unknown value is sent through the normal listing/filtering path, where it matches nothing and becomes a successful empty response.

The fix should centralize the accepted status set and add tests for a single invalid value and a mixed comma-separated value such as `running,definitely-invalid`.

### Regression evidence

- Case: `30j-invalid-ps-status`
- Command: `agent-compose --json ps --all --status definitely-invalid`
- Observed exit code: `0`
- Observed JSON field: `"sandboxes": null`
- No stdout/stderr diagnostic identified the invalid status.

fuchencong:
## Summary

- centralize the canonical sandbox status set and reject unknown values instead of treating them as empty-result filters
- validate `ps` and `sandbox ls` before client creation, validate V2 `ListSandboxes` before storage access, and validate Scheduler `vmStatus` at the shared JSON adapter boundary
- keep storage permissive for internal callers and historical records; unfiltered `--all` can still display unknown persisted statuses
- restrict local `sandbox prune --status` input to the domain-safe `stopped` and `failed` states
- document the accepted status contract without changing protobuf wire types

Invalid CLI status values now return usage exit code 2 with empty stdout and do not send daemon requests. Direct API callers receive Connect `InvalidArgument`.

Fixes #455

## Testing

- `GOFLAGS=-buildvcs=false go test ./pkg/model ./pkg/agentcompose/api ./pkg/agentcompose/adapters ./cmd/agent-compose -count=1`
- `task docs:build`

Full test coverage summary: Unit 75.13%, Integration 60.76%, E2E 60.57%, Combined 79.15%.

`GOFLAGS=-buildvcs=false` is the documented local linked-worktree compatibility setting. Runtime JavaScript dependencies were installed from the committed lockfile before the final full test run.

## Checklist

### DIFF
--- a/cmd/agent-compose/cli_project_display.go
+++ b/cmd/agent-compose/cli_project_display.go
@@ -6,6 +6,8 @@ import (
 	"sort"
 	"strings"
 	"text/tabwriter"
+
+	domain "agent-compose/pkg/model"
 )
 
 type composeDisplayChangeBuilder struct {
@@ -55,7 +57,11 @@ func composePSStatusFilter(options composePSOptions) (map[string]bool, error) {
 		if value == "" {
 			continue
 		}
-		result[value] = true
+		status, err := domain.NormalizeSandboxVMStatus(value)
+		if err != nil {
+			return nil, fmt.Errorf("invalid --status %q: expected %s", value, composePSSupportedStatusExpectation())
+		}
+		result[strings.ToLower(status)] = true
 	}
 	if len(result) > 0 {
 		return result, nil
@@ -64,6 +70,14 @@ func composePSStatusFilter(options composePSOptions) (map[string]bool, error) {
 		return nil, nil
 	}
 	return map[string]bool{"running": true}, nil
+}
+
+func composePSSupportedStatusExpectation() string {
+	statuses := domain.SupportedSandboxVMStatuses()
+	for index := range statuses {
+		statuses[index] = strings.ToLower(statuses[index])
+	}
+	return strings.Join(statuses[:len(statuses)-1], ", ") + ", or " + statuses[len(statuses)-1]
 }
 
 func composePSStatusValues(filter map[string]bool) []string {
--- a/cmd/agent-compose/cli_ps_command.go
+++ b/cmd/agent-compose/cli_ps_command.go
@@ -18,7 +18,7 @@ func newCLISandboxListCommand(cli *cliOptions, use, short string, options *compo
 		},
 	}
 	cmd.Flags().BoolVarP(&options.All, "all", "a", false, "Show current project sandboxes in all statuses")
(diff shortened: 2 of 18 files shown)


eetoc:
@monkeyscan review 下这个PR

eetoc (approved):
LGTM

PR #464, opened 2026-07-27
fix(ci): rerun protobuf gate after approval

kingfs:
## Summary

  - Separate Protobuf compatibility checks from review events so different event types no longer create conflicting required checks.
  - Remove the redundant `protobuf-breaking-approved` label requirement. A maintainer approval for the current PR head commit is now sufficient.
  - Add a review listener that automatically reruns the original Protobuf compatibility workflow after an approval, change request, or review dismissal.
  - Share a concurrency group between the compatibility and review workflows so early reviews wait for the initial Buf check to finish.
  - Keep the review listener isolated from PR code: it does not checkout or execute PR contents and only requests an Actions rerun.
  - Update the CI contract test to enforce the new approval flow and permission boundary.

  The resulting flow is:

  1. Push a PR containing a breaking Protobuf change.
  2. The compatibility check reports that maintainer approval is required.
  3. A maintainer approves the current head commit.
  4. The original required check is rerun automatically and passes.
  5. A new push invalidates the previous approval because the head SHA changes.

  ## Testing

  - `ruby -e 'require "yaml"; ARGV.each { |path| YAML.parse_file(path) }' .github/workflows/proto-compatibility.yml .github/workflows/proto-compatibility-
  review.yml`
  - `./scripts/tests/test-proto-breaking-ci-contract.sh`
  - `task test:scripts`
  - `task lint`
  - Verified the workflow-run lookup against the GitHub Actions API using an existing Protobuf compatibility run.
  - `task test` was run after `task prepare`, but did not complete successfully because existing `SandboxRPCBridge` tests require
  `AGENT_COMPOSE_RUNTIME_BASE_URL` in the local environment. The workflow contract, deployment checks, and unrelated packages completed before that
  environment-dependent failure.

  ## Checklist

### DIFF
--- a/.github/workflows/proto-compatibility-review.yml
+++ b/.github/workflows/proto-compatibility-review.yml
@@ -1,0 +1,38 @@
+name: Refresh Protobuf compatibility approval
+
+on:
+  pull_request_review:
+    types: [submitted, dismissed]
+
+permissions:
+  actions: write
+  contents: read
+  pull-requests: read
+
+# Share the required workflow's group so a review submitted while Buf is still
+# running waits for that run to finish before requesting a new attempt.
+concurrency:
+  group: proto-compatibility-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+
+jobs:
+  refresh-proto-breaking-approval:
+    name: Refresh Protobuf compatibility approval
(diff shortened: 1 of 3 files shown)


eetoc (approved):
LGTM

PR #489, opened 2026-07-28
feat(sandbox): make stopped runtime retention configurable

### LINKED ISSUE #442: feat(sandbox): make stopped runtime retention configurable
## Problem

All runtime drivers currently retain stopped runtime state so a sandbox can resume with its private writable state:

- Docker retains the stopped container (`AutoRemove: false`) and its writable layer.
- BoxLite stops but retains the box/root disk (`auto_remove = 0`).
- Microsandbox stops but retains the sandbox and its private qcow2 rootfs overlay.

This is the correct persistence default, but disposable workloads can accumulate stopped containers, boxes, and overlays. Users can only remove the whole sandbox or run explicit prune; there is no declarative way to keep sandbox metadata and persistent mounts while releasing the stopped runtime.

Out-of-band Docker/runtime pruning is unsafe because it bypasses agent-compose ownership and leaves a sandbox that appears resumable even though its private writable state is gone.

## Proposal

Add a driver-independent sandbox lifecycle policy to each agent:

```yaml
agents:
  worker:
    sandbox:
      stopped_runtime_policy: retain # retain | remove
```

The exact block name can follow the compose schema conventions, but the policy should not be named after Docker `AutoRemove`.

Semantics:

- `retain` : preserve current behavior. Stop retains the driver runtime and private writable state; resume restarts the same runtime.
- `remove`(default): after a confirmed stop, explicitly remove only the driver runtime and its private writable state. Keep the sandbox record, events, logs, workspace, declared volumes, and other durable mounts. Resume recreates a runtime from the snapshotted image/config and mounts. Data that existed only in the private writable layer/root disk is intentionally lost.

Driver mapping:

| Driver | Retained by `retain` | Released by `remove` |
| --- | --- | --- |
| Docker | stopped container and writable layer | container |
| BoxLite | stopped box and private root disk | box/runtime-owned private disk |
| Microsandbox | stopped sandbox and private qcow2 rootfs overlay | sandbox plus private overlay/sidecar |

Keep runtime auto-remove disabled. The lifecycle coordinator should explicitly release the runtime after stop so failures are observable, state transitions are recoverable, and ownership checks remain enforced.

## State model

1. Parse and normalize `stopped_runtime_policy` in `pkg/compose`; default missing values to `retain`.
2. Snapshot the effective policy and required recreation inputs into sandbox runtime metadata at creation. Later compose edits affect only new sandboxes.
3. Add an explicit persisted release intent/state that distinguishes deliberate runtime removal from unexpected runtime loss.
4. Under the existing per-sandbox lifecycle lock: persist release intent, stop the runtime, release only the runtime, then clear its ID/ref after confirmed removal or not-found.
5. Resume may recreate only when persisted state proves removal was intentional. A retained sandbox whose runtime disappeared must continue to fail rather than silently losing its writable state.
6. If stop succeeds but release fails, keep the sandbox stopped, retain the runtime reference, record an event/error, and allow an idempotent retry.
7. Startup recovery must finish an interrupted intentional release without deleting unrelated resources.

This needs a focused runtime-release capability separate from full sandbox removal. Full `sandbox rm`/`sandbox prune --force` must continue through the existing staged removal coordinator and delete metadata/accessories as well.

Ownership should remain separated: policy/defaulting in the sandbox domain/compose boundary, lifecycle decisions in `pkg/sessions`, and driver-specific release mechanics in `pkg/driver`.

## Compatibility

- Default is `retain`; existing compose projects and sandboxes behave unchanged.
- Legacy sandbox metadata without the field decodes as `retain`.
- Existing stopped sandboxes are not retroactively switched when project configuration changes.
- `sandbox rm`, `sandbox prune`, and `WORKSPACE_CLEANUP_TTL` keep their current independent semantics.
- The effective policy and whether the runtime is retained/released should be visible in JSON/verbose sandbox inspection and lifecycle events.

## Alternatives considered

- Expose Docker `auto_remove`: Docker-specific, races control-plane persistence, and does not cover BoxLite/Microsandbox.
- Automatically run backend prune: bypasses sandbox ownership and produces inconsistent metadata.
- Global environment variable only: cannot express different retention needs per agent/project. A global default could be a later addition.
- TTL-only cleanup: useful as a follow-up, but does not provide deterministic stop behavior. A later stopped-runtime TTL can reuse the same safe release capability.

## Acceptance criteria

- Compose accepts only `retain` and `remove`, defaults to `retain`, and English/Chinese YAML manuals are updated.
- Each driver passes a lifecycle test proving `retain` preserves private writable data and the runtime identity across stop/resume.
- Each driver passes a lifecycle test proving `remove` releases its private runtime resource; resume creates a new runtime and preserves declared durable mounts but not private writable-layer data.
- Deliberate release permits recreation; unexpected loss under `retain` does not.
- Release is idempotent for not-found, retryable after failure/interruption, and serialized against concurrent stop/resume/remove.
- Managed-resource inventory does not report successfully released runtime resources as orphan residue.
- Focused unit/integration tests cover parsing, legacy defaults, state transitions, recovery, and driver mappings; lifecycle concurrency tests run with `go test -race`.

winterfx:
## Summary

Closes #442.

- add the driver-independent `sandbox.stopped_runtime_policy` with backward-compatible `retain` default and per-sandbox snapshots
- persist recoverable `retained`, `release_pending`, and `released` lifecycle state and expose it through sandbox inspection and events
- release only runtime-private resources for `remove`, while preserving sandbox metadata, workspace, logs, events, and durable mounts
- serialize stop, release, resume, removal, and startup recovery under the sandbox lifecycle boundary
- centralize the stop-required decision from VM state and start-attempt fences, and roll back unowned runtime starts to a clean retryable state
- document lifecycle states and stop/release ordering in the English and Chinese manuals

## Testing

- `go test ./pkg/model ./pkg/compose ./pkg/storage/sandboxstore ./pkg/sandboxes ./pkg/agentcompose/adapters ./pkg/agentcompose/api ./pkg/agentcompose/app ./pkg/projects ./pkg/runs`
- `go test -race ./pkg/sandboxes ./pkg/agentcompose/adapters ./pkg/agentcompose/app ./pkg/runs`
- `task lint`
- `task build`
- `task docs:build`
- `TMPDIR=/private/tmp ./scripts/test-coverage.sh`
  - unit: 74.73%
  - integration: 60.67%
  - E2E: 60.17%
  - combined: 78.77%
- local OrbStack Docker lifecycle smoke for both policies
  - `retain` reused the exact runtime ID and preserved private-layer and workspace data
  - `remove` created a new runtime ID, discarded private-layer data, and preserved workspace data

`task test` could not complete its deploy preflight on macOS because the deploy scripts currently assume Linux/GNU paths and tools (`/var`, `sha256sum`, and GNU `sed`). The test and coverage portion was run directly with `TMPDIR=/private/tmp` as listed above.

## Checklist


eetoc (approved):
LGTM

PR #495, opened 2026-07-28
fix(llms): normalize custom v1 endpoints

sxyzzzzzzz:
## Summary

- recognize OpenAI-compatible endpoints with arbitrary path prefixes ending in `/v1`
- append `chat/completions` or `responses` consistently for env-backed providers
- add regression coverage for custom-prefixed Qwen-compatible endpoints and the env-default provider resolution path

## Problem

`NormalizeAPIBaseURL` removes an explicit protocol suffix before rebuilding the endpoint. `NormalizeAPIEndpointForProtocol` only rebuilt exact `/v1` and `/openai/v1` paths, so a base such as `/models/qwen/v1` was requested directly and returned 404.

## Verification

- `go test ./pkg/llms`
- `task lint` (0 issues)
- `task build`
- `task test` progressed through script contracts and failed only in installer tests because macOS maps `/var` through a symlink (`refusing install path with symlink component: /var`); unrelated to this change

### DIFF
--- a/pkg/llms/config_endpoint_test.go
+++ b/pkg/llms/config_endpoint_test.go
@@ -1,0 +1,43 @@
+package llms
+
+import "testing"
+
+func TestNormalizeAPIEndpointForProtocolCustomV1Prefix(t *testing.T) {
+	tests := []struct {
+		name     string
+		protocol string
+		want     string
+	}{
+		{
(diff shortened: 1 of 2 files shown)


eetoc (approved):
LGTM

PR #498, opened 2026-07-28
fix(api): decode persisted scheduler specs

eetoc:
## Summary

- decode persisted project scheduler snapshots through the normalized compose representation
- map scheduler and trigger policies back to protobuf enums at the API boundary
- use the shared decoder for scheduler details, manual invocation, and declared trigger lookup

## Problem

Project reconciliation persists `compose.NormalizedSchedulerSpec` as JSON, where policies and trigger kinds are strings such as `new`, `skip`, and `event`. After the protobuf fields changed to enums, the API attempted to decode those snapshots directly with `encoding/json` into `agentcomposev2.SchedulerSpec`. `GetScheduler` and scheduler execution then failed with `cannot unmarshal string into Go struct field SchedulerSpec.sandbox_policy`.

## Tests

- `go test ./pkg/agentcompose/api`
- `task lint`
- `task build`
- `task test`
- real daemon regression with existing data: `GetScheduler` returned 200
- real UI regression: scheduler run `516ab8a0a61d88f5e18a7739f3f07a99508f58ae7bd9b4bc748c1555f92eaa8f` succeeded and its Agent executed the Shell marker `PROTOCOL_AGENT_SHELL_OK`

## Checklist

- [x] Regression tests added
- [x] No protobuf or generated source changed
- [x] No configuration or documentation changes required

### DIFF
--- a/pkg/agentcompose/api/project_handler.go
+++ b/pkg/agentcompose/api/project_handler.go
@@ -266,8 +266,8 @@ func runtimeTriggerSpec(trigger domain.SchedulerTrigger) *agentcomposev2.Trigger
 }
 
 func declaredTriggerSpec(scheduler domain.ProjectSchedulerRecord, triggerID string) *agentcomposev2.TriggerSpec {
-	var spec agentcomposev2.SchedulerSpec
-	if json.Unmarshal([]byte(scheduler.SpecJSON), &spec) != nil {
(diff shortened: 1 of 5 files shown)


eetoc:
@monkeyscan review下

fuchencong (approved):
LGTM

PR #511, opened 2026-07-29
fix(docker): support configurable GitHub mirror

sxyzzzzzzz:
## Summary

- add a GITHUB_MIRROR Docker build argument with https://github.com as the default
- use the configured mirror for BoxLite and Microsandbox release downloads
- add an image CI contract check to prevent hard-coded native runtime release URLs

## Validation

- bash scripts/tests/test-image-ci-contract.sh
- docker buildx build --check --build-arg REGISTRY_MIRROR=registry-mirrors.dev.in.chaitin.net -f Dockerfile .
- verified the internal mirror returns HTTP 200 for representative BoxLite and Microsandbox release assets


eetoc (approved):
LGTM

PR #521, opened 2026-07-29
fix(runs): finalize runs after sandbox cleanup

innomentats:
## Problem

A project run could previously be persisted as terminal before its sandbox cleanup completed. This created several inconsistent and unsafe states:

- a run could report succeeded, failed, or canceled while its sandbox was still consuming compute resources
- a stop or removal failure happened after the terminal result was already committed, so retrying cleanup could not preserve the invariant between run state and sandbox state
- a daemon crash between terminal persistence and cleanup could permanently leave an active sandbox behind
- startup recovery reconstructed a generic failed result instead of resuming the exact completion that had already been selected
- StopRun wrote CANCELED directly, racing with the execution result and bypassing cleanup
- a sandbox awaiting cleanup could be reused by another run while the old cleanup was retrying

The core issue was that terminal-state persistence and sandbox lifecycle management were two independent operations with the terminal write happening first.

## Fix approach

This change establishes one invariant: a run may enter a terminal state only after its configured completion cleanup has succeeded. KEEP_RUNNING is the explicit exception because it requires no cleanup.

Completion is split into three durable phases:

1. Stage the exact first completion result and stable final agent/result events in SQLite. Keep the run RUNNING and do not write a terminal status event.
2. Execute the derived cleanup action idempotently and retry failures in the background with bounded backoff.
3. After cleanup succeeds, atomically commit the terminal run fields, append the unique terminal status event, clear cleanup_error, and delete the completion journal.

This ordering closes the daemon-crash windows while preserving the exact original result through retries and restarts. Competing completion or cancellation requests use first-writer-wins semantics and cannot overwrite the staged result.

### Cleanup semantics

- STOP_ON_COMPLETION: stop the associated sandbox before finalizing the run.
- REMOVE_ON_COMPLETION: fully remove a sandbox created by this run; a reused sandbox is stopped instead.
- KEEP_RUNNING: perform no cleanup and finalize immediately.
- no associated sandbox: perform no cleanup and finalize immediately.

Already-stopped or missing sandboxes are treated as successful stop outcomes. Removal is also idempotent when both the sandbox and its deletion journal are already gone, covering a crash after external deletion but before the terminal transaction.

### Durable completion and recovery

SQLite schema v10 adds persisted cleanup policy and sandbox ownership fields plus project_run_completion, a recovery journal containing the target terminal state, exact serialized transition, cleanup action, retry state, last error, and next attempt time.

Two completion workers process due journals. Cleanup failures leave the run RUNNING, populate cleanup_error, and do not create a terminal status event. Retries use 1s, 2s, 4s, 8s, 16s, and 30s delays, then remain at 30s.

On daemon startup, sandbox runtime state is reconciled first. Existing completion journals resume their exact staged result. Legacy PENDING or RUNNING runs without a journal receive an interrupted FAILED completion, then follow the same cleanup and finalization path. Completion recovery starts before schedulers so a new run cannot claim a sandbox that is still being cleaned up.

### Execution and cancellation

All success, execution failure, cancellation, preparation failure, sandbox startup failure, command, agent, streaming, detached, and attach paths now complete through the same completion manager.

RunSupervisor registers unary, streaming, detached, and attach executions with context.WithCancelCause. StopRun now requests cancellation with the user reason instead of directly writing CANCELED. The execution stages the cancellation result, applies cleanup, and commits CANCELED only after cleanup succeeds. Repeated StopRun calls remain idempotent.

### Sandbox safety

The run and sandbox association is persisted before a new runtime is started or an existing sandbox is resumed. A binding failure prevents runtime startup; a newly created sandbox is rolled back through the removal path.

Explicit and sticky reuse reject a sandbox that has a pending completion journal, preventing cleanup retries from racing with new execution.

## Changes made

- added SQLite v10 migration and standalone migrator schema/checksum updates
- added cleanup policy/action domain normalization and persistence
- added runs.CompletionManager with durable staging, retry workers, idempotent cleanup, and atomic finalization
- persisted sandbox bindings before runtime start/resume
- routed all run completion paths through the completion manager
- changed StopRun to asynchronous supervised cancellation
- added daemon startup recovery and shutdown ordering for completion work
- blocked reuse of pending-completion sandboxes
- updated English and Chinese CLI documentation plus the design document
- added unit, persistence, recovery, cancellation, and sandbox-binding coverage
- added an opt-in real Docker hard-restart E2E that verifies an interrupted run becomes FAILED only after its container stops consuming active compute resources

## User-visible behavior

- RunAgent, StreamAgentRun, and foreground CLI runs continue waiting while cleanup is retrying. Client cancellation ends the wait but background completion continues.
- StartAgentRun still returns immediately after submission.
- StopRun may return stop_requested=true while the run is still RUNNING. The run becomes CANCELED after cleanup succeeds.
- cleanup_error explains the latest cleanup failure while the run remains recoverable.
- CompletedAt and DurationMs are calculated when terminal state is committed, so duration includes cleanup retry time.
- permanent driver failures intentionally leave the run RUNNING until the runtime is repaired or removed.

## Testing

- task lint
- task build
- task test
  - unit: 74.89%
  - integration: 60.88%
  - E2E: 60.33%
  - combined: 79.07%
- go test -race ./pkg/runs ./pkg/agentcompose/app ./pkg/sandboxes -timeout 300s
- focused tests for runs, app, API, sandboxes, SQLite/configstore, projects, and standalone migrator
- task docs:build
- task test:e2e:docker-run-completion-recovery

## Checklist

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -94,6 +94,19 @@ it before creating and validating a date-partitioned sandbox of its own.
 This focused task is intentionally absent from `task test` and GitHub Actions
 because GitHub-hosted CI does not provide its prebuilt guest-image prerequisite.
 
+Interrupted run cleanup has an opt-in hard-restart Docker E2E:
+
+```bash
+task image:agent-compose-guest
+task test:e2e:docker-run-completion-recovery
+```
+
+Set `AGENT_COMPOSE_E2E_RUN_COMPLETION_IMAGE` to use another compatible local
+guest image. The test starts a long-running command through the public API,
+hard-kills the host daemon, restarts it against the same data root, and verifies
+that the interrupted run becomes `FAILED` only after its Docker sandbox no
+longer consumes active compute resources.
+
 The daemon retention cleanup has a separate opt-in real Docker E2E:
 
 ```bash
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -243,6 +243,22 @@ tasks:
           GOCACHE={{.GO_BUILD_CACHE_DIR}} \
           {{.GO_TOOLCHAIN_ENV}} go test ./test/e2e -run '^TestE2EDockerFileWorkspaceResumePreservesState$' -count=1 -v
 
+  test:e2e:docker-run-completion-recovery:
+    desc: Run the opt-in host-daemon Docker hard-restart run completion E2E test
+    dir: '{{.TASKFILE_DIR}}'
+    deps: ['build:agent-compose']
+    cmds:
+      - |
+        image="${AGENT_COMPOSE_E2E_RUN_COMPLETION_IMAGE:-agent-compose-guest:latest}"
+        if ! docker image inspect "$image" >/dev/null 2>&1; then
+          echo "Docker run-completion image $image is unavailable; run 'task image:agent-compose-guest' or set AGENT_COMPOSE_E2E_RUN_COMPLETION_IMAGE" >&2
+          exit 1
+        fi
+        AGENT_COMPOSE_E2E_BINARY="$PWD/build/agent-compose" \
+          AGENT_COMPOSE_E2E_RUN_COMPLETION_IMAGE="$image" \
+          GOCACHE={{.GO_BUILD_CACHE_DIR}} \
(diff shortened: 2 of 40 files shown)


innomentats:
@monkeyscan review this pr

rokku-c:
@monkeyscan 重新评审

PR #525, opened 2026-07-30
feat(events): add efficient webhook event queries

eetoc:
## Summary

- Add efficient, extensible HTTP event queries through `GET /api/events?source=webhook&view=summary`, `GET /api/events/topics?source=webhook`, and `GET /api/events/:event_id/trace`.
- Keep existing `/api/events`, `/runs`, `/sandboxes`, and `/sessions` behavior compatible without adding protobuf or Connect APIs.
- Bound trace expansion at 1,000 events, batch SQLite sandbox projections, and keep trace results available when optional sandbox enrichment fails.
- Preserve loader-era event history by matching both stored `scheduler` and legacy `loader` source values in full, summary, and topic queries.
- Add schema v11 indexes and update the webhook design documentation.

## Testing

- `go test ./pkg/model ./pkg/storage/configstore ./pkg/events/webhooks`
- `task lint`
- `task build`
- `task test` was not run because the existing full test task is blocked on macOS by `/var` realpath assumptions and GNU `sed -i` usage that is incompatible with BSD sed.

## Checklist


rokku-c:
@monkeyscan 重新评审

fuchencong (approved):
LGTM

PR #540, opened 2026-08-02
fix(llms): inject max_output_tokens into proxied requests to stop silent long-output truncation

trtyr:
## Problem (business-impacting)

When a model is served via an **OpenAI-compatible gateway** (DeepSeek, vLLM, Ollama, third-party proxies), agent-compose's codex sandboxes **silently truncate long agent outputs** mid-response, and the truncated run returns an **empty/missing result**. This is structural (not transient), so retries do not recover it, and it breaks real workloads — e.g. in production SOC pipelines, verbose analysis nodes (a multi-host EDR process-tree analysis) deterministically fail.

## Root cause

Two issues stack:

1. **codex never sends `max_output_tokens`** in its Responses API requests (openai/codex#36180 — the wire struct, ModelInfo, and config all lack the field).
2. agent-compose's protocol bridge therefore falls back to a **small default `max_completion_tokens` (4096)** on the translated chat-completions request, and **upstream proxies honor `max_completion_tokens`** over `max_tokens`.

Result: the model is capped at ~4096 output tokens regardless of the model's real capability, the Responses API returns *"Incomplete response returned, reason: max_output_tokens"*, and the codex runner decodes no result payload → the agent run fails. (Confirmed by capturing the facade's outgoing body: `max_completion_tokens: 4096` even though the model can emit far more — a direct streaming probe returned 2.9 MB with `max_tokens: 65536`.)

Note: a codex `model_catalog_json` does **not** fix this — per openai/codex#36180, `ModelInfo` has no `max_output_tokens` field, so codex ignores any limit declared in a custom catalog.

## Fix

When the operator sets **`LLM_MAX_OUTPUT_TOKENS`**, the runtime LLM facade injects the limit into every proxied upstream request:

- `max_output_tokens` for the responses protocol
- **both** `max_tokens` and `max_completion_tokens` for chat completions (proxies honor `max_completion_tokens`, so both must be overridden to beat the bridge's 4096 default)

Opt-in via env — default behavior is unchanged for built-in models.

### Configuration

```env
LLM_MAX_OUTPUT_TOKENS=65536
```

## Verification

End-to-end against DeepSeek behind an OpenAI-compatible gateway (baizhi cloud). Same 3-host EDR analysis card across fix iterations:

| Build | Failed nodes | Result |
|---|---|---|
| upstream (no fix) | 1 | EDR_Tree truncated → run failed |
| catalog-only attempt | 1 | catalog ignored (codex#36180) |
| facade injects `max_tokens` only | 2 | proxy honors `max_completion_tokens`=4096, still truncated |
| **this PR (injects both fields)** | **0** | **full 22 KB report, zero truncation** |

A direct streaming probe to the same gateway returned 2.9 MB with `max_tokens: 65536`, confirming the model/gateway can emit long outputs once the facade injects the limit.

## Changes

- `pkg/agentcompose/proxy/runtime_llm.go` — `injectMaxOutputTokens` rewrites the proxied upstream body in both the transparent and protocol-bridged paths.
- `pkg/agentcompose/app/app.go` — passes `config.LLMMaxOutputTokens` into the facade handler options.
- `pkg/config/config.go` + `pkg/config/codex_runtime.go` — read `LLM_MAX_OUTPUT_TOKENS`.
- `.env.example` — documents the variable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -149,6 +149,13 @@ DEFAULT_IMAGE=chaitin/agent-compose-guest:latest
 # LLM_API_KEY=
 # OPENAI_API_KEY=
 # LLM_MODEL=
+# Maximum output tokens the runtime LLM facade injects into each proxied
+# upstream request. codex does not send max_output_tokens itself, and the
+# protocol bridge defaults max_completion_tokens to a small value (e.g. 4096),
+# so proxies that honor that field silently truncate long agent outputs.
+# Set this for models served via OpenAI-compatible gateways (DeepSeek, vLLM,
+# Ollama, third-party proxies) to stop mid-response truncation.
+# LLM_MAX_OUTPUT_TOKENS=65536
 # Bounds direct LLM calls and the facade's wait for upstream response headers.
 # Established streaming responses do not have a total duration limit.
 # LLM_TIMEOUT=60s
--- a/demo/agent-compose.yml
+++ b/demo/agent-compose.yml
@@ -1,8 +1,0 @@
-name: demo
-
-agents:
-  hello:
-    provider: codex
-    image: chaitin/agent-compose-guest:latest
-    driver:
-      docker: {}
--- a/demo/agent-compose.yml
+++ b/demo/agent-compose.yml
@@ -1,0 +1,8 @@
+name: demo
+
+agents:
+  hello:
+    provider: codex
+    image: chaitin/agent-compose-guest:latest
+    driver:
+      docker: {}
--- a/pkg/agentcompose/app/app.go
+++ b/pkg/agentcompose/app/app.go
@@ -606,7 +606,8 @@ func registerRuntimeLLMFacadeRoutes(app *echo.Echo, di do.Injector) {
 		ResolveTarget: func(ctx context.Context, requestedModel, providerID string) (llms.ResolvedTarget, error) {
 			return llms.ResolveRuntimeLLMTarget(ctx, config, configDB, requestedModel, providerID)
 		},
-		Client: proxy.NewRuntimeLLMHTTPClient(config.LLMTimeout),
+		Client:          proxy.NewRuntimeLLMHTTPClient(config.LLMTimeout),
+		MaxOutputTokens: config.LLMMaxOutputTokens,
 	})
 }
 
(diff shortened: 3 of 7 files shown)


winterfx:
Hi @trtyr ,

Thanks for tracking down this structural truncation issue and for putting together the PR. Identifying the protocol bridge default of 4096, connecting it to missing Codex output limits, and validating the fix end to end against a long DeepSeek response is very useful work.

I found one protocol-specific regression in the current implementation at d977fecb. injectMaxOutputTokens treats every upstream protocol other than OpenAI Responses as OpenAI Chat Completions:

    if upstreamProtocol == protocolbridge.ProtocolOpenAIResponses {
        obj["max_output_tokens"] = maxTokens
    } else {
        obj["max_tokens"] = maxTokens
        obj["max_completion_tokens"] = maxTokens
    }

ProtocolAnthropicMessages also enters this else branch. With LLM_MAX_OUTPUT_TOKENS=65536, the emitted Anthropic request contains:

    {
      "model": "claude-test",
      "messages": [],
      "max_tokens": 65536,
      "max_completion_tokens": 65536
    }

Anthropic Messages uses max_tokens and does not define max_completion_tokens. I reproduced this on the PR branch with a strict Anthropic Messages schema; the request was rejected with:

    json: unknown field "max_completion_tokens"

This affects both transparent Messages to Messages forwarding and converted Responses or Chat to Messages requests, because both paths call injectMaxOutputTokens. The existing proxy package tests still pass, so this case currently has no regression coverage.

A protocol-specific switch would avoid sending cross-protocol fields:

    switch upstreamProtocol {
    case protocolbridge.ProtocolOpenAIResponses:
        obj["max_output_tokens"] = maxTokens
    case protocolbridge.ProtocolOpenAIChat:
        obj["max_completion_tokens"] = maxTokens
    case protocolbridge.ProtocolAnthropicMessages:
        obj["max_tokens"] = maxTokens
    }

It would also be helpful to add request-body tests for all three upstream protocols. Thanks again for finding and addressing the original long-output failure.

winterfx (approved):

eetoc (approved):
LGTM

PR #546, opened 2026-08-03
fix(ci): prevent non-release tags updating latest

eetoc:
## Summary

- disable docker/metadata-action automatic latest tag generation in the manifest merge job
- keep stable latest and archlinux tags controlled exclusively by the existing v-prefixed release-tag condition
- add an image CI contract regression check for the disabled automatic latest flavor

This prevents tags such as runtime-v0.8.0 from overwriting the published latest daemon and guest manifests.

## Testing

- `./scripts/tests/test-image-ci-contract.sh`
- `bash -n scripts/tests/test-image-ci-contract.sh`
- `git diff --check`

Not run: `task test:deploy` because Task and Go are unavailable in the current environment. ShellCheck and actionlint are also unavailable.

## Checklist

- [x] Documentation updated when behavior or configuration changed. No public documentation change is required for this CI-only correction.

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -321,6 +321,9 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ${{ env.IMAGE_PREFIX }}/${{ matrix.image }}
+          # Stable tags are selected explicitly below. The action's automatic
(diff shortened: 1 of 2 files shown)


winterfx (approved):
LGTM

PR #547, opened 2026-08-03
fix(image): enable capability proxy by default

eetoc:
## Summary

- enable the internal gRPC capability proxy by default in published and local daemon images
- listen on 0.0.0.0:7420 and inject agent-compose:7420 as the guest-reachable target
- lock both image defaults in the Docker deployment contract test

## Security

- the proxy continues to require the per-sandbox CAP_TOKEN and validates capset membership
- guest authorization is stripped before forwarding; the configured OctoBus token is used upstream
- Docker Compose does not publish port 7420 to the host

## Testing

- PASS: git diff --check
- PASS: static contract checks for both Dockerfiles
- NOT COMPLETED: go test ./cmd/agent-compose -run capability deployment contract tests; the clean Go container remained blocked fetching dependencies for more than 13 minutes

## Documentation

- no public documentation changes; this PR only supplies image-level defaults and leaves docker-compose.yml unchanged


winterfx (approved):
lgtm

PR #561, opened 2026-08-04
fix(ci): add debug logging for DingTalk release notification

eetoc:
## Summary

Add debug logging to diagnose why the DingTalk notify workflow fails at the LLM translation step.

## Changes

- Print release tag, body length, and request body size before LLM call
- Print raw LLM response (first 4000 bytes) after curl
- Include raw response in error message when `choices[0].message.content` extraction fails

Relates to: https://github.com/chaitin/agent-compose/actions/runs/30923584539


winterfx (approved):
lgtm

PR #569, opened 2026-08-06
fix(cli): probe h2c capability before attach streams (issue #413)

EasonW3300:
## Summary
  - Add a side-effect-free pre-attach h2c capability probe: before opening an AttachAgentRun / AttachExec bidi stream, issue `GET {baseURL}/api/version` over a raw h2c transport (matching the existing health-check URL construction) and propagate only the HTTP/2-vs-HTTP/1 transport mismatch, so an HTTP/1-only
  reverse proxy is reported with exit code 3 and guidance instead of failing mid-attach.
  - Skip probing entirely for unix sockets and non-`http://` base URLs, keeping the default local daemon path and the https path byte-for-byte unchanged.
  - Extract `newDaemonH2CTransport` from `newDaemonBaseRoundTripper` for reuse by the probe (behavior-equivalent refactor).
  - Extend `isAttachHTTP2TransportMismatch` to also recognize the x/net/http2 `http2: client conn could not be established` race variant, where the connection is torn down before the first stream is registered.
  - Add unit tests T1–T9 covering probe skip cases, HTTP/1-only mismatch detection, h2c success, transient/timeout pass-through, entry-function exit-code mapping, and real-stream pass-through.

## Testing
  - `go vet ./cmd/agent-compose/...` — clean
  - `go build ./...` — clean
  - `go test ./cmd/agent-compose/...` — all pass, incl. probe tests under `-race`
  - `go test -cover ./cmd/agent-compose/` — 80.4% of statements
  - `go test ./...` — `cmd/agent-compose` passes; `pkg/compose` and `pkg/volumes` have pre-existing macOS path-symlink / timezone failures, confirmed unchanged on a clean tree via `git stash`

## Checklist

### DIFF
--- a/cmd/agent-compose/cli_client.go
+++ b/cmd/agent-compose/cli_client.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"agent-compose/proto/agentcompose/v2/agentcomposev2connect"
+	"context"
 	"os"
 	"strings"
 	"time"
@@ -26,6 +27,12 @@ type cliServiceClients struct {
 	cache    agentcomposev2connect.CacheServiceClient
 	volume   agentcomposev2connect.VolumeServiceClient
 	sandbox  agentcomposev2connect.SandboxServiceClient
+
+	// attachBidiProbe is a side-effect-free pre-attach h2c capability check,
+	// or nil when probing does not apply (unix socket, https). It runs before
+	// any AttachAgentRun/AttachExec bidi stream is opened so an HTTP/1-only
+	// proxy is reported before resources are prepared.
+	attachBidiProbe func(context.Context) error
 }
 
 func newCLIServiceClients(cli cliOptions) (cliServiceClients, error) {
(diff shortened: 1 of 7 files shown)


eetoc:
@monkeyscan review 这个pr

eetoc:
I do not think this is the desired product direction.

The CLI is expected to connect directly to the daemon through its Unix socket or an HTTP/2-capable daemon endpoint. An HTTP/1 reverse proxy is not a supported CLI topology. Adding a pre-attach `GET /api/version` h2c probe makes every plain-HTTP TCP attach, including the normal direct-daemon path, pay for detection and error-classification logic that exists only for an unsupported deployment mode.

I verified the implementation against a real daemon and a real Docker guest. The probe behaves as described: direct h2c `AttachAgentRun` and `AttachExec` succeed, while an HTTP/1-only Nginx endpoint fails early with exit code 3 and does not create resources or execute the command. That proves the implementation works according to this PR, but it does not make the proxy topology part of the intended contract.

Please remove the pre-attach probe and instead document/enforce the endpoint boundary: CLI traffic must go directly to a Unix socket or HTTP/2-capable endpoint; UI/general HTTP ingress must not be reused as the CLI attach endpoint. If remote CLI access is required, it should use native HTTP/2, TLS HTTP/2, or an SSH/VPN tunnel rather than compatibility logic for an HTTP/1 proxy.

One additional concern is that the probe only applies to `http://` and is skipped for `https://`, so it does not establish a coherent transport-capability contract for remote endpoints.

PR #572, opened 2026-08-08
fix(scheduler): preserve sticky command runtime contracts

zhuanyongxigua:
## Summary

- rebuild the command-scoped startup and selected-provider LLM facade environment whenever a Scheduler command uses a loaded or reused Sandbox
- persist `scheduler.command.started` with the linked Sandbox before executing a trigger command, replacing the extra `scheduler.shell("true")` prelink
- derive sticky compatibility from the selected Agent and Scheduler's effective Sandbox inputs instead of the enclosing Project revision

The original failure came from relying on transient facade environment fields that are intentionally absent from persisted Sandbox metadata. A later command or a sticky reuse could load the Sandbox without the startup Anthropic/OpenAI facade environment, producing an endpoint/token mismatch. The command path now reconstructs that managed environment for the child process and owns the tokens created for that command.

For a reused sticky Sandbox there may be no new `sandbox.created` or `sandbox.resumed` event. The new pre-execution Scheduler event makes the SchedulerRun-to-Sandbox association durable without running a separate `true` command. Failure to persist the association prevents the command from starting.

Sticky bindings created with the older hash schema may be replaced once after upgrading because the old hash included the Project revision. Subsequent changes to unrelated Agents in the same Project no longer invalidate the replacement binding; changes to effective image, driver, environment, volumes, workspace, capsets, or Agent configuration still do.

## Testing

Passed on `a447c8ce`:

- `GOTOOLCHAIN=go1.26.2 go test ./pkg/agentcompose/adapters ./pkg/llms/runtimefacade ./pkg/schedulers ./pkg/storage/configstore -count=1`
- targeted `-race -count=1` tests for command facade reconstruction/token ownership, pre-command Run-to-Sandbox linking/failure behavior, and sticky hash semantics
- `GOTOOLCHAIN=go1.26.2 task lint`
- `GOTOOLCHAIN=go1.26.2 task build`
- `GOTOOLCHAIN=go1.26.2 task --force docs:build`
- `git diff --check origin/main...HEAD`

`GOTOOLCHAIN=go1.26.2 task test` reaches `test:deploy` and then fails in the installer suite because macOS exposes `/var` as a symlink. The same `go -C cmd/installer test ./...` failure was reproduced on the unchanged `origin/main@b874e660`; this PR does not modify installer code.

## Scope

This Draft intentionally excludes:

- webhook parent Event lineage, fallback/root trace changes, and webhook header/idempotency hardening
- opt-in `.git/index.lock` cleanup during Sandbox resume
- the separate late-`AddCell` STOPPED-state persistence fix
- AgentKit loader changes

Before marking ready, the exact PR build still needs the dev180 sticky E2E: two sequential runs reuse one Sandbox, the second run reaches the facade without a 401, the active run exposes its Sandbox ID before command completion, unrelated Agent changes do not replace the binding, and effective Sandbox configuration changes do.

## Checklist

### DIFF
--- a/pkg/schedulers/run_host.go
+++ b/pkg/schedulers/run_host.go
@@ -289,6 +289,9 @@ func (h *RuntimeHost) Command(ctx context.Context, request domain.SchedulerComma
 		_ = h.addLinkedSchedulerEvent(ctx, eventType, "info", "scheduler command sandbox ready", map[string]any{"sandboxId": session.Summary.ID}, session.Summary.ID, "", "")
 	}
 	h.trackCommandSession(session.Summary.ID, cleanupSession)
+	if err := h.persistCommandSandboxLink(ctx, session.Summary.ID); err != nil {
+		return domain.SchedulerCommandResult{}, err
+	}
 
 	result, err := h.deps.CommandExecutor.ExecuteSchedulerCommand(ctx, session, request)
 	if err != nil {
@@ -301,6 +304,19 @@ func (h *RuntimeHost) Command(ctx context.Context, request domain.SchedulerComma
 	}
 	_ = h.addLinkedSchedulerEvent(ctx, "scheduler.command.completed", level, firstHostNonEmpty(result.Output, result.Stdout, result.Stderr, "scheduler command completed"), CommandEventPayload(request, result), result.SandboxID, result.CellID, "")
 	return result, nil
+}
+
+func (h *RuntimeHost) persistCommandSandboxLink(ctx context.Context, sandboxID string) error {
+	if h.execution.Kind != ExecutionKindTrigger {
+		return nil
+	}
+	if h.deps.Events == nil {
+		return fmt.Errorf("persist scheduler command sandbox link: scheduler event recorder is unavailable")
+	}
+	if err := h.addLinkedSchedulerEvent(ctx, "scheduler.command.started", "info", "scheduler command started", map[string]any{"sandboxId": sandboxID}, sandboxID, "", ""); err != nil {
+		return fmt.Errorf("persist scheduler command sandbox link: %w", err)
+	}
+	return nil
 }
 
 func (h *RuntimeHost) LLM(ctx context.Context, prompt string, request domain.SchedulerLLMRequest) (domain.SchedulerLLMResult, error) {
--- a/pkg/schedulers/run_host.go
+++ b/pkg/schedulers/run_host.go
@@ -289,7 +289,7 @@ func (h *RuntimeHost) Command(ctx context.Context, request domain.SchedulerComma
 		_ = h.addLinkedSchedulerEvent(ctx, eventType, "info", "scheduler command sandbox ready", map[string]any{"sandboxId": session.Summary.ID}, session.Summary.ID, "", "")
 	}
 	h.trackCommandSession(session.Summary.ID, cleanupSession)
-	if err := h.persistCommandSandboxLink(ctx, session.Summary.ID); err != nil {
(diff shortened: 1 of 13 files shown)


zhuanyongxigua:
@monkeyscan review pr

winterfx:
The overall direction looks right: rebuilding command-scoped facade tokens/environment for every command fixes sticky sandbox reuse without persisting transient state in the sandbox. The token ownership and cleanup changes also look reasonable.

One part seems broader than necessary:

```go
runtimeCommandRequest.Env =
    mergeSchedulerCommandRuntimeEnv(request.Env, managedFacadeEnv)
```

`request.Env` comes from explicit command options such as `scheduler.shell(..., { env })`. The guest runtime already defines it as overriding the outer runtime environment. Replacing those values here changes that contract and also writes raw facade tokens into `command-request.json`.

The code path also shows why preserving this precedence does not break sticky reuse:

1. `schedulerCommandRequestFromOptions` only populates `request.Env` from the explicit `env` command option; sticky sandbox reuse does not populate it from persisted sandbox state.
2. `prepareSchedulerCommandLLMFacadeEnv` rebuilds fresh managed facade values for every command and puts them on the per-command `execSession.RuntimeEnvItems`.
3. `BuildSandboxExecEnv` passes those fresh values to the outer runtime process. The guest then applies `request.env` last only when the caller explicitly supplied a conflicting value.

Therefore, with the existing contract, ordinary sticky commands still receive the newly generated facade token. The stale provider values in the test are manually inserted into `SchedulerCommandRequest.Env`; they are not state produced by sticky reuse.

A minimal regression assertion can be added to the existing executor test:

```go
if got := runtimeRequest.Env["OPENAI_API_KEY"]; got != "request-upstream-openai-key" {
    t.Fatalf("OPENAI_API_KEY = %q, want explicit request value", got)
}
```

This fails on the current PR because the generated facade token replaces the explicit request value.

I also tested removing only this merge. Fresh facade tokens still worked with the same sticky sandbox across sequential reuse, stop/resume, daemon reconstruction, and real facade HTTP requests. So the sticky fix appears to require rebuilding `execSession.RuntimeEnvItems`, but not overriding `request.Env`.

Could we keep the managed facade environment on `ExecSpec`/`RuntimeEnvItems` and pass the original request unchanged to `RuntimeCommandRequestPayload`?

winterfx:
@monkeyscan review这个pr

yorkew-east8:
@monkeyscan 重新评审一下这个 PR

pkg/schedulers/run_host.go
```diff
@@ -303,6 +306,36 @@ func (h *RuntimeHost) Command(ctx context.Context, request domain.SchedulerComma
 	return result, nil
 }
 
+func (h *RuntimeHost) persistCommandSandboxLink(ctx context.Context, request domain.SchedulerCommandRequest, sandboxID string) error {
+	if h.execution.Kind != ExecutionKindTrigger {
+		return nil
+	}
+	if h.deps.Events == nil {
```
kingfs: This introduces a compatibility regression for trigger commands when the event recorder is not configured. Elsewhere in this host, addSchedulerEvent/addSchedulerEventRecord deliberately treat Events == nil as an optional dependency and return nil; Command, Agent, and LLM therefore still execute without event persistence. With this new check, every trigger Command now returns an error before ExecuteSchedulerCommand runs, even though no persistence attempt could fail.\n\nPlease either keep this path optional (for example, return nil when Events == nil) or update the dependency contract and all callers/tests to make Events mandatory. A regression test covering ExecutionKindTrigger with Events == nil would make the intended behavior explicit.

winterfx (approved):
LGTM

eetoc (approved):

PR #578, opened 2026-08-11
fix(config): sync env default LLM providers

eetoc:
## Summary

- fix stale persisted credentials after `SettingsService.UpdateGlobalEnv` changes the global environment
- resolve OpenAI and Anthropic credentials independently using the existing `global_env → process environment → daemon config` precedence
- reconcile all enabled `env_default` providers in the matching provider family, including custom environment-managed provider IDs
- keep `system` and `session_env` providers unchanged
- update the global environment and managed provider credentials atomically in one SQLite transaction
- preserve Anthropic API-key and auth-token header semantics

An `env_default` provider is environment-managed by definition. Therefore, when its provider family is reconciled, all enabled `env_default` providers in that family receive the same effective environment credential. A custom provider that intentionally requires an independent credential should use a different scope rather than `env_default`.

## Testing

- `go test ./...`
- `go vet ./pkg/llms ./pkg/storage/configstore ./pkg/agentcompose/api`
- `git diff --check`
- reproduced the stale persisted-provider credential behavior on the base branch
- verified the same daemon/API/SQLite flow on this branch updates persisted provider credentials
- verified authenticated OpenAI-compatible and Anthropic-compatible requests using the updated credentials
- verified custom `env_default` providers are reconciled
- verified `system` and `session_env` providers remain unchanged
- verified removing global credentials falls back to the next configured source

## Checklist

- [x] Documentation updated when behavior or configuration changed. No documentation change is required because this restores the existing environment-provider behavior and does not add configuration or API surface.

### DIFF
--- a/pkg/llms/provider_reconcile.go
+++ b/pkg/llms/provider_reconcile.go
@@ -1,0 +1,55 @@
+package llms
+
+import (
+	domain "agent-compose/pkg/model"
+)
+
+// EnvDefaultProviderCredentials contains credentials derived from global
+// environment values for the automatically managed default providers.
+type EnvDefaultProviderCredentials struct {
+	OpenAIAPIKey        string
+	AnthropicAPIKey     string
+	AnthropicAuthHeader string
+	AnthropicAuthScheme string
+}
+
+// ResolveEnvDefaultProviderCredentials applies the same credential precedence
+// used by environment-provider bootstrap.
+func ResolveEnvDefaultProviderCredentials(items []domain.SandboxEnvVar) EnvDefaultProviderCredentials {
+	return ResolveEnvDefaultProviderCredentialsFromLayers(items)
+}
+
+// ResolveEnvDefaultProviderCredentialsFromLayers resolves credentials from
+// layers in precedence order. Each layer is exhausted before the next layer is
+// considered, preserving the bootstrap source-major precedence.
+func ResolveEnvDefaultProviderCredentialsFromLayers(layers ...[]domain.SandboxEnvVar) EnvDefaultProviderCredentials {
+	openAIKey := resolveFamilyAPIKey(layers, "LLM_API_KEY", "OPENAI_API_KEY")
+	anthropic := resolveFamilyAnthropicCredential(layers)
+	return EnvDefaultProviderCredentials{OpenAIAPIKey: openAIKey, AnthropicAPIKey: anthropic.apiKey, AnthropicAuthHeader: anthropic.authHeader, AnthropicAuthScheme: anthropic.authScheme}
+}
+
+func resolveFamilyAPIKey(layers [][]domain.SandboxEnvVar, keys ...string) string {
+	for _, layer := range layers {
+		if value := firstNonEmptyTrimmed(envValues(layer, keys...)...); value != "" {
+			return value
+		}
+	}
+	return ""
+}
+
+func resolveFamilyAnthropicCredential(layers [][]domain.SandboxEnvVar) anthropicCredential {
+	for _, layer := range layers {
+		if credential, ok := anthropicCredentialFromItems(layer); ok {
+			return credential
+		}
+	}
+	return anthropicCredential{authHeader: "x-api-key"}
+}
+
+func envValues(items []domain.SandboxEnvVar, keys ...string) []string {
+	values := make([]string, 0, len(keys))
+	for _, key := range keys {
+		values = append(values, EnvItemValue(items, key))
+	}
+	return values
+}
--- a/pkg/storage/configstore/core_store.go
+++ b/pkg/storage/configstore/core_store.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"agent-compose/pkg/llms"
 	domain "agent-compose/pkg/model"
 )
 
@@ -96,6 +97,15 @@ func (s *coreStore) ListGlobalEnv(ctx context.Context) ([]domain.SandboxEnvVar, 
 }
 
 func (s *coreStore) ReplaceGlobalEnv(ctx context.Context, items []domain.SandboxEnvVar) ([]domain.SandboxEnvVar, error) {
+	credentials := llms.ResolveEnvDefaultProviderCredentials(items)
+	return s.replaceGlobalEnv(ctx, items, &credentials)
+}
+
+func (s *coreStore) ReplaceGlobalEnvWithProviderCredentials(ctx context.Context, items []domain.SandboxEnvVar, credentials llms.EnvDefaultProviderCredentials) ([]domain.SandboxEnvVar, error) {
+	return s.replaceGlobalEnv(ctx, items, &credentials)
+}
+
+func (s *coreStore) replaceGlobalEnv(ctx context.Context, items []domain.SandboxEnvVar, credentials *llms.EnvDefaultProviderCredentials) ([]domain.SandboxEnvVar, error) {
 	normalized := domain.NormalizeEnvItems(items)
 	tx, err := s.db.BeginTx(ctx, nil)
 	if err != nil {
@@ -110,6 +120,9 @@ func (s *coreStore) ReplaceGlobalEnv(ctx context.Context, items []domain.Sandbox
 		if _, err := tx.ExecContext(ctx, `INSERT INTO global_env(name, value, secret, updated_at) VALUES(?, ?, ?, ?)`, item.Name, item.Value, BoolToInt(item.Secret), time.Now().UTC().Unix()); err != nil {
 			return nil, fmt.Errorf("insert global env %s: %w", item.Name, err)
 		}
+	}
+	if err := syncEnvDefaultProviderCredentials(ctx, tx, credentials); err != nil {
+		return nil, err
 	}
 	if err := tx.Commit(); err != nil {
 		return nil, fmt.Errorf("commit global env tx: %w", err)
--- a/pkg/storage/configstore/global_env_provider_sync.go
+++ b/pkg/storage/configstore/global_env_provider_sync.go
@@ -1,0 +1,24 @@
+package configstore
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"agent-compose/pkg/llms"
+)
+
+func syncEnvDefaultProviderCredentials(ctx context.Context, tx *sql.Tx, resolved *llms.EnvDefaultProviderCredentials) error {
+	if resolved == nil {
+		return nil
+	}
+	now := time.Now().UTC().Unix()
+	if _, err := tx.ExecContext(ctx, `UPDATE llm_provider SET api_key = ?, updated_at = ? WHERE provider_type = ? AND scope = ? AND api_key != ?`, resolved.OpenAIAPIKey, now, llms.ProviderFamilyOpenAI, llms.ProviderScopeEnvDefault, resolved.OpenAIAPIKey); err != nil {
+		return fmt.Errorf("sync default OpenAI provider credentials: %w", err)
+	}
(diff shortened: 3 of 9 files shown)


pkg/storage/configstore/global_env_provider_sync.go
```diff
@@ -0,0 +1,25 @@
...
+	"database/sql"
+	"fmt"
+
+	"agent-compose/pkg/llms"
+	domain "agent-compose/pkg/model"
+)
+
+func syncEnvDefaultProviderCredentials(ctx context.Context, tx *sql.Tx, previous, current []domain.SandboxEnvVar) error {
+	credentials := llms.ResolveEnvDefaultProviderCredentials(current)
+	if llms.EnvDefaultOpenAIInputsChanged(previous, current) {
```
winterfx: **correctness:** This hardcodes `WHERE id = 'default'` / `WHERE id = 'anthropic'`, so env-default OpenAI-compatible providers created under any other id never get their API key refreshed on rotation.

Failure scenario: `pkg/llms/custom_openai_facade_target.go:32` calls `EnsureOpenAIEnvProvider(..., providerID, providerID, ProviderScopeEnvDefault, ...)` with a caller-chosen custom `providerID` (reached from the OpenCode/Pi facade's default branch for a custom OpenAI-compatible provider reference). Once that `env_default`-scoped `llm_provider` row exists and is enabled, `HasEnabledLLMProviderID` short-circuits future resolution and it's never re-bootstrapped. An admin later rotates `OPENAI_API_KEY` via `ReplaceGlobalEnv`; this only updates `id IN ('default','anthropic')`, so the custom-id row silently keeps the old API key forever — even though propagating rotated credentials to env-default providers is this PR's stated purpose.

winterfx (commented):
Review findings from an automated pass — flagging the credential-rotation gaps as the priority since they undercut this PR's stated goal.

pkg/storage/configstore/global_env_provider_sync.go
```diff
@@ -0,0 +1,25 @@
...
+		if _, err := tx.ExecContext(ctx, `UPDATE llm_provider SET api_key = ? WHERE id = ? AND provider_type = ? AND scope = ?`, credentials.OpenAIAPIKey, llms.ProviderIDDefaultOpenAI, llms.ProviderFamilyOpenAI, llms.ProviderScopeEnvDefault); err != nil {
```
winterfx: **correctness:** The UPDATE filters on `id AND provider_type AND scope`, even though `id` is the table's sole PRIMARY KEY (`llm_provider.id TEXT PRIMARY KEY`). If a row's `scope`/`provider_type` ever diverges from the expected constant while its `id` stays the same, this silently matches zero rows with no error.

Combined with the neighboring finding: an admin rotating credentials via `ReplaceGlobalEnv` would get no error or log, but the provider keeps its stale `api_key` — a second, harder-to-notice failure mode on top of the id-scoping gap.

pkg/llms/provider_reconcile.go
```diff
@@ -0,0 +1,64 @@
...
+type EnvDefaultProviderCredentials struct {
+	OpenAIAPIKey        string
+	AnthropicAPIKey     string
+	AnthropicAuthHeader string
+	AnthropicAuthScheme string
+}
+
+// ResolveEnvDefaultProviderCredentials applies the same credential precedence
+// used by environment-provider bootstrap.
+func ResolveEnvDefaultProviderCredentials(items []domain.SandboxEnvVar) EnvDefaultProviderCredentials {
```
winterfx, line 18: **simplification:** `ResolveEnvDefaultProviderCredentials` re-derives the Anthropic credential precedence (`ANTHROPIC_API_KEY > ANTHROPIC_AUTH_TOKEN > LLM_API_KEY`, with matching header/scheme) via an inline `lookup` + `AnthropicProviderAuthFromLookup`, alongside the existing `anthropicCredentialFromValues` (`pkg/llms/anthropic_credential.go`) which encodes the same precedence independently.

Risk: the same rule now lives in at least two independently-maintained implementations. A future change to the precedence order or header/scheme mapping applied to one path but not the other would make env-default credential rotation silently diverge from every other Anthropic credential resolution path in the codebase, with no compiler error to catch the mismatch.

pkg/storage/configstore/core_store.go
```diff
@@ -95,14 +96,26 @@ func (s *coreStore) ListGlobalEnv(ctx context.Context) ([]domain.SandboxEnvVar,
 	return items, nil
 }
 
+func (s *ConfigStore) ReplaceGlobalEnv(ctx context.Context, items []domain.SandboxEnvVar) ([]domain.SandboxEnvVar, error) {
+	return replaceGlobalEnv(ctx, s.db, items, true)
+}
+
 func (s *coreStore) ReplaceGlobalEnv(ctx context.Context, items []domain.SandboxEnvVar) ([]domain.SandboxEnvVar, error) {
```
winterfx, line 99: **dead-code:** `*coreStore` is only ever constructed embedded inside `*ConfigStore` (via `FromDB`), and `*ConfigStore` defines its own `ReplaceGlobalEnv` that always shadows this promoted embedded method via Go's method-resolution rules — so the `syncProviders=false` branch here can never actually run through any real code path.

Not an active bug, but a maintenance trap: a future refactor that stores or exposes a bare `*coreStore` (e.g. a lighter test harness) would silently inherit the non-syncing behavior — skipping provider credential sync entirely — with no compiler warning, since the two near-identical methods look interchangeable at a glance.

pkg/storage/configstore/core_store.go
```diff
@@ -111,12 +124,39 @@ func (s *coreStore) ReplaceGlobalEnv(ctx context.Context, items []domain.Sandbox
 			return nil, fmt.Errorf("insert global env %s: %w", item.Name, err)
 		}
 	}
+	if syncProviders && llms.EnvDefaultProviderInputsChanged(previous, normalized) {
```
winterfx: **simplification:** `EnvDefaultProviderInputsChanged` is computed once here as a gate, then `syncEnvDefaultProviderCredentials` immediately recomputes the same thing per-family (`EnvDefaultOpenAIInputsChanged` / `EnvDefaultAnthropicInputsChanged`) against the exact same `previous`/`normalized` values.

No functional bug — just doubles the same key-by-key string comparisons on every `ReplaceGlobalEnv` call for zero new information, and forces a reader to check two separate functions to understand whether a sync actually happens.

pkg/storage/configstore/core_store.go
```diff
@@ -95,14 +96,26 @@ func (s *coreStore) ListGlobalEnv(ctx context.Context) ([]domain.SandboxEnvVar,
...
+func replaceGlobalEnv(ctx context.Context, db *sql.DB, items []domain.SandboxEnvVar, syncProviders bool) ([]domain.SandboxEnvVar, error) {
 	normalized := domain.NormalizeEnvItems(items)
-	tx, err := s.db.BeginTx(ctx, nil)
+	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return nil, fmt.Errorf("begin global env tx: %w", err)
 	}
 	defer func() { _ = tx.Rollback() }()
 
+	previous, err := listGlobalEnvTx(ctx, tx)
```
winterfx: **efficiency:** `listGlobalEnvTx` runs unconditionally inside every transaction, even when `syncProviders` is false, purely to feed a change-detection check that's immediately discarded in that case.

Low absolute cost given `ReplaceGlobalEnv` is an infrequent admin operation, but it's unconditional work (an extra full-table SELECT + row scan) that could be gated behind `if syncProviders`.

eetoc:
Addressed the review findings in commit `220dacb1`:

- reconcile all `env_default` providers by provider family, including custom OpenAI-compatible provider IDs
- remove the redundant `ConfigStore` shadow method and non-syncing branch
- reuse the existing Anthropic credential precedence implementation
- retain atomic SQLite updates and add coverage for custom OpenAI provider rotation

Focused tests pass:
`go test ./pkg/llms ./pkg/storage/configstore`

eetoc:
@monkeyscan review 下整个PR

winterfx:
关于最新一轮提交(220dacb1)的一个小遗留问题(不在本次 diff 范围内,GitHub 没法挂 inline comment,所以写在这里):

**`pkg/llms/provider_bootstrap.go:147` `AnthropicProviderAuthFromLookup` 现在是死代码**

这次提交把 `ResolveEnvDefaultProviderCredentials`(`pkg/llms/provider_reconcile.go`)改成直接调用 `anthropicCredentialFromItems`,不再走 `AnthropicProviderAuthFromLookup`。搜了一下,这个函数现在全仓库没有任何调用方了。

不是运行时 bug(它是导出函数,编译器不会报),但确实是这次清理漏掉的死代码,而且是 `anthropicCredentialFromValues` 优先级逻辑的一份过期重复实现,以后有人找"Anthropic auth header 优先级在哪判断"容易踩到这个过时分支。建议顺手删掉。

eetoc:
Removed the unused `AnthropicProviderAuthFromLookup` helper from `pkg/llms/provider_bootstrap.go` as requested.

Focused tests still pass:
`go test ./pkg/llms ./pkg/storage/configstore`

pkg/storage/configstore/global_env_provider_sync.go
```diff
@@ -0,0 +1,25 @@
...
+	"fmt"
+
+	"agent-compose/pkg/llms"
+	domain "agent-compose/pkg/model"
+)
+
+func syncEnvDefaultProviderCredentials(ctx context.Context, tx *sql.Tx, previous, current []domain.SandboxEnvVar) error {
+	credentials := llms.ResolveEnvDefaultProviderCredentials(current)
+	if llms.EnvDefaultOpenAIInputsChanged(previous, current) {
+		if _, err := tx.ExecContext(ctx, `UPDATE llm_provider SET api_key = ? WHERE provider_type = ? AND scope = ?`, credentials.OpenAIAPIKey, llms.ProviderFamilyOpenAI, llms.ProviderScopeEnvDefault); err != nil {
```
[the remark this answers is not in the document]
eetoc: 感谢指出这个边界。我们检查了当前仓库的完整赋值链路：custom provider ID 只来自 OpenCode/Pi 的模型命名空间，custom provider 的 API key 仍由 DefaultLLMEnvProviderLookup 从 global_env、进程环境或 daemon 配置读取；当前没有独立的 env_default provider 凭据配置 API 或页面。显式独立凭据属于 system/session_env provider，而这两类不会被本次同步覆盖。

因此，对当前支持的语义，所有 env_default provider 都是环境凭据的持久化快照，global_env 轮换时应统一同步。若增加 previous api_key 条件，反而会让历史不一致的 env_default provider 继续保留旧 key，无法修复本 PR 的目标。我们已增加 custom provider ID 跟随环境凭据轮换的回归测试。

winterfx (dismissed):

eetoc:
Added effective credential fallback handling in commit `652e0ccd`.

`UpdateGlobalEnv` now resolves provider credentials using the same source-layer precedence as bootstrap:

`global_env → daemon process environment → daemon config`

The resolved credentials are passed into the same SQLite transaction, so removing a global key correctly falls back to process/config credentials instead of clearing the persisted provider key. Added regression coverage for this fallback path.

Focused tests pass:
`go test ./pkg/llms ./pkg/storage/configstore ./pkg/agentcompose/api`

eetoc:
Additional validation: `go test ./...` passes using the internal Go toolchain/registry setup.

The fallback implementation also preserves source-major precedence for Anthropic credentials, so a higher-priority global auth token is not accidentally displaced by a lower-priority process API key.

pkg/storage/configstore/global_env_provider_sync.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"time"
+
+	"agent-compose/pkg/llms"
+)
+
+func syncEnvDefaultProviderCredentials(ctx context.Context, tx *sql.Tx, resolved *llms.EnvDefaultProviderCredentials) error {
+	if resolved == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
eetoc: Thanks for flagging this. The broadcast is intentional for this change: `env_default` is the environment-managed scope, so every enabled provider in the matching family is resolved from the effective global environment credential. This is required for custom provider IDs created by the runtime facade; restricting the update to the two well-known IDs would reintroduce stale credentials for those providers. Providers that need an independent credential must use a non-`env_default` scope. The PR description now documents this contract explicitly, and the regression tests cover custom env-default rotation plus preservation of system/session-scoped providers.

PR #588, opened 2026-08-14
fix(api): remove scheduler list N+1 hydration

### LINKED ISSUE #586: [Performance]: remove N+1 scheduler hydration from ListSchedulers
## Summary

`ProjectService/ListSchedulers` currently performs an expensive `GetScheduler` hydration for every scheduler returned by the page. On Devboard, a project with 29 automations already makes this path noticeably slow.

This is a focused follow-up to #536 and covers only the `ListSchedulers` read path.

## Steps to reproduce

1. Start the daemon with a project containing multiple schedulers.
2. Accumulate scheduler run/event history.
3. Call `agentcompose.v2.ProjectService/ListSchedulers` with a page containing multiple schedulers.
4. Observe that latency and SQL work grow with the number of schedulers returned.

## Actual behavior

The handler performs:

```text
ListProjectSchedulersPage             # 1 SQL
CountProjectSchedulers                # 1 SQL
for each scheduler:
  effectiveSchedulerEnabled
    GetScheduler
      loadSchedulerDefinition         # 1 SQL + full revision JSON parse
      listSchedulerTriggers           # 1 SQL
      hydrateSchedulerSummaryCounts   # 1 SQL with scalar aggregates
```

Returning `N` schedulers therefore causes approximately `2 + 3N` SQL statements, plus `N` parses of project revision JSON.

The per-item hydration is redundant:

- `ListProjectSchedulersPage` already returns `project_scheduler.enabled`.
- The page query already returns run count and latest run time.
- `ListSchedulers` only consumes `definition.Summary.Enabled` from the fully hydrated scheduler.
- Full hydration additionally loads triggers and event/run statistics that this RPC does not return.

The page SQL also calculates `COUNT(scheduler_run)` and `MAX(started_at)` as correlated subqueries for every returned scheduler.

Relevant code:

- `pkg/agentcompose/api/project_handler.go`: `ListSchedulers`, `effectiveSchedulerEnabled`
- `pkg/storage/configstore/project_store.go`: `ListProjectSchedulersPage`
- `pkg/storage/configstore/scheduler_definition_store.go`: `loadSchedulerDefinition`
- `pkg/storage/configstore/scheduler_store.go`: `hydrateSchedulerSummaryCounts`

## Expected behavior

Returning one scheduler page should use a constant number of SQL statements with respect to page size and should not load full scheduler definitions per item.

## Proposed direction

- Treat `ProjectSchedulerRecord.Enabled` returned by the page query as authoritative.
- Remove the per-item `GetScheduler` call from `ListSchedulers`.
- Compute run count and latest run time once with a grouped aggregate joined to the requested page, rather than correlated subqueries.
- Preserve the existing response fields, filtering, ordering, pagination, and exact total behavior.

## Acceptance criteria

- [ ] `ListSchedulers` makes no per-item `GetScheduler` calls.
- [ ] SQL statement count remains constant as page size increases.
- [ ] Run count and latest run time are computed without per-row correlated history scans.
- [ ] Existing enabled-state, filtering, ordering, pagination, and total semantics are preserved.
- [ ] A regression test fails if per-item scheduler hydration is reintroduced.
- [ ] A scale-oriented test covers at least 100 schedulers with scheduler run history.

## Version / runtime

- Current repository state as of 2026-08-14
- Runtime driver: not runtime-related

## Related

- Focused implementation issue split from #536 after approval to address the scheduler hot path first.

winterfx:
## Summary

- use the scheduler page record as the authoritative enabled state
- remove per-item `GetScheduler` hydration from `ListSchedulers`
- aggregate scheduler run counts and latest run times once for the requested page
- add a regression assertion against per-item hydration and a 120-scheduler history test

Fixes #586

## Tests

- `task lint`
- `task build`
- `go test ./pkg/agentcompose/api ./pkg/storage/configstore`
- `go test ./cmd/... ./pkg/... ./proto/health/v1 ./proto/health/v1/healthv1connect ./proto/agentcompose/v2 ./proto/agentcompose/v2/agentcomposev2connect` *(scheduler-related packages pass; existing unrelated failures remain in `pkg/compose` timezone validation and `pkg/volumes` macOS `/var` vs `/private/var` path expectations)*
- `task test` *(blocked in installer deployment tests because macOS resolves `/var` through the `/private/var` symlink, producing `refusing install path with symlink component: /var`)*

## Checklist

- [x] Focused regression coverage added
- [x] Scale coverage includes at least 100 schedulers with run history
- [x] No public API or documentation changes

### DIFF
--- a/pkg/agentcompose/api/handler_coverage_test.go
+++ b/pkg/agentcompose/api/handler_coverage_test.go
@@ -767,21 +767,32 @@ func TestProjectAndRunHandlersStoreBackedWorkflows(t *testing.T) {
 	if err != nil || scheduler.Msg.GetScheduler().GetSchedulerId() != "scheduler-1" || scheduler.Msg.GetScheduler().GetDisplayName() != "每日巡检" || scheduler.Msg.GetSpec().GetScript() != "run()" || scheduler.Msg.GetSpec().GetDescription() != "每天汇总巡检结果" {
 		t.Fatalf("GetScheduler resp=%#v err=%v", scheduler, err)
 	}
+	// GetScheduler hydrates scheduler-1 as disabled, diverging from the
+	// project_scheduler page row (Enabled: true), so the calls below can
+	// prove ListSchedulers treats the page row as authoritative instead of
+	// re-hydrating each item via GetScheduler.
+	getSchedulerCallsBeforeList := store.getSchedulerCalls
 	firstSchedulers, err := projectHandler.ListSchedulers(ctx, connect.NewRequest(&agentcomposev2.ListSchedulersRequest{Limit: 1}))
 	if err != nil || len(firstSchedulers.Msg.GetSchedulers()) != 1 || firstSchedulers.Msg.GetTotal() != 2 {
 		t.Fatalf("ListSchedulers first page=%#v err=%v", firstSchedulers, err)
 	}
-	if summary := firstSchedulers.Msg.GetSchedulers()[0]; summary.GetEnabled() || summary.GetRunCount() != 3 || !summary.GetLatestRunAt().AsTime().Equal(time.Unix(10, 0)) || summary.GetLastError() != "failed" || summary.GetDisplayName() != "每日巡检" || summary.GetDescription() != "每天汇总巡检结果" {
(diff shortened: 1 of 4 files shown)


eetoc (approved):
LGTM

Issue #565, opened 2026-08-05
design(storage): reduce large log and output payloads persisted in SQLite


winterfx:
## Problem

Agent Compose currently persists several kinds of log-like content and execution output in `DATA_ROOT/data.db`.

Persisting structured run metadata and audit events is useful for querying, pagination, restart recovery, and transactional consistency. However, some records also contain complete command, agent, LLM, or RPC output. In several cases, the same content is stored both in SQLite and in artifact files, or duplicated across multiple database columns.

This issue proposes auditing the current persistence boundary and evaluating whether large or unbounded output can be moved to artifacts while retaining queryable metadata and compatibility.

### Scheduler events

Trigger-based scheduler executions persist structured events in `scheduler_event`, including:

- `scheduler.log`
- `scheduler.agent.completed` / `scheduler.agent.failed`
- `scheduler.command.completed` / `scheduler.command.failed`
- `scheduler.llm.completed` / `scheduler.llm.failed`
- `scheduler.sandbox.rpc.completed` / `scheduler.sandbox.rpc.failed`
- scheduler run lifecycle events
- scheduler sandbox lifecycle events
- event publication and compatibility warning events

The table stores both `message` and `payload_json`.

Some event types may persist large or duplicated content:

- Agent completion events use the agent result as both the event message and payload. The payload may contain `text`, `output`, and `finalText`.
- LLM completion events store response text in both the message and payload.
- Command completion events store the complete command output in `message`.
- Successful sandbox RPC events store complete request and response JSON in `payload_json`.
- Scheduler completion results are stored in both `scheduler_run.result_json` and the `scheduler.run.completed` event payload.

Only trigger-run hosts configure the persistent scheduler event recorder. Foreground `scheduler invoke` executions do not persist outer scheduler logs.

### Project runs

`project_run` persists:

- `prompt`
- `output`
- `error` and `error_stack`
- `result_json`
- `logs_path`
- `artifacts_dir`

Raw command and agent output is also stored in artifact files such as `stdout.txt`, `stderr.txt`, and `output.txt`.

As a result, `project_run.output` may duplicate the corresponding output artifact. The database copy is currently used by run detail responses and non-follow `logs` behavior, while follow mode reads the file referenced by `logs_path`.

### Project run events

`project_run_event` persists semantic run history:

- user messages
- final agent messages
- terminal status and result metadata

This is not a raw stdout/stderr transcript. It supports ordered and paginated conversation history and can be committed transactionally with run state.

[...]

PR #596, opened 2026-08-18
feat(schedulers): stop persisting command output in scheduler_event.message

winterfx:
## Summary
Code implementation of the plan documented (and tracked separately) in `docs/design/scheduler_event_storage_design.md` (issue #565; supersedes #594, which was design-doc-only). `scheduler.command.completed` events no longer write full command output into `scheduler_event.message`; the daemon reconstructs it on read from the sandbox cell artifact when serving `scheduler logs`.

- `pkg/schedulers/run_host.go` / `payload.go`: `scheduler.command.completed` now writes `message=""` unconditionally; `payload_json` gains `outputBytes`.
- `pkg/schedulers/event_message.go` (new): `ResolveEventMessage(ctx, event, sandboxDirs)`. A non-empty DB message (row written before this change) is returned as-is with zero I/O — the DB and the artifact were written from the same in-memory result at the same time, so there's nothing for a fresh read to correct. An empty message (new row) reads the sandbox cell artifact, checking `output.txt` first and only falling back to `stdout.txt`/`stderr.txt` when needed; if the artifact is unavailable it synthesizes an `outputBytes`-sized placeholder. `EventMessageNeedsArtifactRead` exposes the same predicate to callers.
- `pkg/agentcompose/api`: wires `ResolveEventMessage` into `ListProjectSchedulerEvents` and `StreamProjectSchedulerEvents` via a shared, bounded-concurrency helper — only events that actually need the artifact read are scheduled onto a worker pool (capped at 8), everything else resolves inline. `ListSchedulerEvents` (UI-only, zero callers) is intentionally left untouched. `ProjectHandler` now takes a `schedulers.SandboxDirResolver`, wired from `*sandboxstore.Store` in `app.go`.
- `cmd/agent-compose/e2e_docker_scheduler_test.go`: switched the docker e2e assertion from `ListSchedulerEvents` (never resolves the artifact) to `ListProjectSchedulerEvents`.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./pkg/schedulers/... ./pkg/agentcompose/...`
- [x] `go test ./pkg/schedulers/... ./pkg/agentcompose/api/... ./pkg/agentcompose/app/...` (including `-race`)
- [x] `go build/vet -tags docker_e2e ./cmd/agent-compose/...`
- [x] Manual end-to-end verification against a locally built daemon (real docker sandbox, real scheduler run): confirmed the `scheduler_event` DB row's `message` is empty while `scheduler logs`/`scheduler logs --json` return the full reconstructed output, and `payload_json` carries `outputBytes` with no `messageTruncated`.
- [ ] `docker_e2e`-tagged `TestSchedulerE2E...` run in CI (needs the tagged runner; only compile-checked locally)

### DIFF
--- a/cmd/agent-compose/e2e_docker_scheduler_test.go
+++ b/cmd/agent-compose/e2e_docker_scheduler_test.go
@@ -284,7 +284,12 @@ func waitForScheduledHelloRun(t *testing.T, projectClient agentcomposev2connect.
 	deadline := time.Now().Add(timeout)
 	var lastErr error
 	for time.Now().Before(deadline) {
-		eventsResp, err := projectClient.ListSchedulerEvents(context.Background(), connect.NewRequest(&agentcomposev2.ListSchedulerEventsRequest{
+		// ListProjectSchedulerEvents (not the UI-only ListSchedulerEvents) is used here
+		// because scheduler.command.completed rows now write an empty DB message and
+		// only ListProjectSchedulerEvents/StreamProjectSchedulerEvents reconstruct the
+		// full text from the sandbox cell artifact (see
+		// docs/design/scheduler_event_storage_design.md §6).
+		eventsResp, err := projectClient.ListProjectSchedulerEvents(context.Background(), connect.NewRequest(&agentcomposev2.ListProjectSchedulerEventsRequest{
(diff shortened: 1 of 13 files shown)


winterfx:
Moved to a fresh branch: superseded by #597.

PR #597, opened 2026-08-18
feat(schedulers): stop persisting command output in scheduler_event.message

winterfx:
## Summary
Code implementation of the plan documented (and tracked separately) in `docs/design/scheduler_event_storage_design.md` (issue #565; supersedes #594 and #596, both closed). `scheduler.command.completed` events no longer write full command output into `scheduler_event.message`; the daemon reconstructs it on read from the sandbox cell artifact when serving `scheduler logs`.

- `pkg/schedulers/run_host.go` / `payload.go`: `scheduler.command.completed` now writes `message=""` unconditionally; `payload_json` gains `outputBytes`.
- `pkg/schedulers/event_message.go` (new): `ResolveEventMessage(ctx, event, sandboxDirs)`. A non-empty DB message (row written before this change) is returned as-is with zero I/O. An empty message (new row) reads the sandbox cell artifact, checking `output.txt` first and only falling back to `stdout.txt`/`stderr.txt` when needed; if the artifact is unavailable it synthesizes an `outputBytes`-sized placeholder. `EventMessageNeedsArtifactRead` exposes the same predicate to callers.
- `pkg/agentcompose/api`: wires `ResolveEventMessage` into `ListProjectSchedulerEvents`, `StreamProjectSchedulerEvents`, and now `ListSchedulerEvents` via a shared, bounded-concurrency `resolveSchedulerEventMessages` helper — only events that actually need the artifact read are scheduled onto a worker pool (capped at 8), everything else resolves inline. `ProjectHandler` now takes a `schedulers.SandboxDirResolver`; the default `NewProjectHandler` exposes it via a new `WithSandboxDirs` chained setter (documented as construction-time-only, matching the existing `WithLifecycleLocks`/`WithRemovalCoordinator`/`WithRunTargetResolver` pattern), while `NewProjectHandlerWithAgentModels` still takes it directly. Wired from `*sandboxstore.Store` in `app.go`.
- `cmd/agent-compose/e2e_docker_scheduler_test.go`: switched the docker e2e assertion from `ListSchedulerEvents` to `ListProjectSchedulerEvents`.

Review follow-ups (addressed in later commits on this branch):
- `ListSchedulerEvents` was initially left unresolved (UI-only, believed to have no callers) — wired into the same resolver so new `command.completed` rows don't show up empty there either.
- `WithSandboxDirs` writes `h.sandboxDirs` without synchronization while requests read it concurrently — doc'd as must-call-before-serving-traffic, consistent with the package's other post-construction injectors; the injection test was rewritten to stop modeling the unsafe "inject after serving" pattern.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./pkg/schedulers/... ./pkg/agentcompose/...`
- [x] `go test ./pkg/schedulers/... ./pkg/agentcompose/api/... ./pkg/agentcompose/app/...` (including `-race`)
- [x] `go build/vet -tags docker_e2e ./cmd/agent-compose/...`
- [x] Manual end-to-end verification against a locally built daemon (real docker sandbox, real scheduler run): confirmed the `scheduler_event` DB row's `message` is empty while `scheduler logs`/`scheduler logs --json` return the full reconstructed output, and `payload_json` carries `outputBytes` with no `messageTruncated`.
- [x] Added `TestListSchedulerEventsResolvesCommandCompletedFromArtifact` and `TestNewProjectHandlerWithSandboxDirsInjectsResolver` to pin the `ListSchedulerEvents` and `WithSandboxDirs` fixes.
- [ ] `docker_e2e`-tagged `TestSchedulerE2E...` run in CI (needs the tagged runner; only compile-checked locally)

### DIFF
--- a/cmd/agent-compose/e2e_docker_scheduler_test.go
+++ b/cmd/agent-compose/e2e_docker_scheduler_test.go
@@ -284,11 +284,11 @@ func waitForScheduledHelloRun(t *testing.T, projectClient agentcomposev2connect.
 	deadline := time.Now().Add(timeout)
 	var lastErr error
 	for time.Now().Before(deadline) {
-		// ListProjectSchedulerEvents (not the UI-only ListSchedulerEvents) is used here
-		// because scheduler.command.completed rows now write an empty DB message and
-		// only ListProjectSchedulerEvents/StreamProjectSchedulerEvents reconstruct the
-		// full text from the sandbox cell artifact (see
-		// docs/design/scheduler_event_storage_design.md §6).
+		// ListProjectSchedulerEvents is used here because it's what `scheduler
+		// logs` actually calls; ListSchedulerEvents (UI-only, unused by any
+		// current caller) also reconstructs command.completed output from the
+		// sandbox cell artifact and is covered separately by
+		// TestListSchedulerEventsResolvesCommandCompletedFromArtifact.
 		eventsResp, err := projectClient.ListProjectSchedulerEvents(context.Background(), connect.NewRequest(&agentcomposev2.ListProjectSchedulerEventsRequest{
 			Project:   &agentcomposev2.ProjectRef{Selector: &agentcomposev2.ProjectRef_ProjectId{ProjectId: scheduler.GetProjectId()}},
 			AgentName: scheduler.GetAgentName(),
--- a/cmd/agent-compose/e2e_docker_scheduler_test.go
+++ b/cmd/agent-compose/e2e_docker_scheduler_test.go
@@ -284,7 +284,12 @@ func waitForScheduledHelloRun(t *testing.T, projectClient agentcomposev2connect.
 	deadline := time.Now().Add(timeout)
 	var lastErr error
 	for time.Now().Before(deadline) {
-		eventsResp, err := projectClient.ListSchedulerEvents(context.Background(), connect.NewRequest(&agentcomposev2.ListSchedulerEventsRequest{
+		// ListProjectSchedulerEvents (not the UI-only ListSchedulerEvents) is used here
+		// because scheduler.command.completed rows now write an empty DB message and
+		// only ListProjectSchedulerEvents/StreamProjectSchedulerEvents reconstruct the
+		// full text from the sandbox cell artifact (see
+		// docs/design/scheduler_event_storage_design.md §6).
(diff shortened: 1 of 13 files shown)


winterfx:
Both findings from the review are accurate, fixed in 996f79b2:

1. **`ListSchedulerEvents` returned an empty message for new `command.completed` rows** — true, it wasn't wired into `resolveSchedulerEventMessages` like the other two RPCs. Now it is; covered by `TestListSchedulerEventsResolvesCommandCompletedFromArtifact`. (In production this endpoint currently has zero callers per the design doc's `agent-compose-ui` audit, so it wasn't a live-traffic regression, but the fix removes the discrepancy regardless.)
2. **`NewProjectHandler` had no way to inject `sandboxDirs`** — true; today it's only called from tests, so nothing in production silently degraded, but it was a real footgun for any future non-test caller. Added `ProjectHandler.WithSandboxDirs`, matching the existing `WithLifecycleLocks`/`WithRemovalCoordinator`/`WithRunTargetResolver` builder pattern already used elsewhere in this package. Covered by `TestNewProjectHandlerWithSandboxDirsInjectsResolver`, which asserts the placeholder-before / reconstructed-after behavior explicitly.

winterfx:
Latest finding is also accurate, fixed in 85fdfa0f:

**\`WithSandboxDirs\` is a post-construction mutable setter racing with concurrent request reads of \`h.sandboxDirs\`** — true. \`ListSchedulerEvents\`/\`ListProjectSchedulerEvents\`/\`StreamProjectSchedulerEvents\` all read it unsynchronized via \`resolveSchedulerEventMessages\`, and the setter wrote it unsynchronized too. Worth noting: this exact shape already exists for this package's other post-construction injectors (\`WithLifecycleLocks\`, \`WithRemovalCoordinator\`, \`WithRunTargetResolver\`) — none of them are synchronized either, because the implicit contract across the package is "call once, immediately after construction, before the handler is registered." I made that contract explicit in \`WithSandboxDirs\`'s doc comment rather than adding \`atomic.Value\` just for this one field, which would've been inconsistent with its siblings for no real safety gain (nothing in production calls \`WithSandboxDirs\` post-construction).

More importantly. the review is right that my own test modeled the exact unsafe pattern (build handler → serve a request → call \`WithSandboxDirs\` → serve another, all on the same handler instance). Rewrote \`TestNewProjectHandlerWithSandboxDirsInjectsResolver\` to build two independent handlers instead, so it no longer demonstrates post-construction mutation as if it were supported usage.

eetoc (approved):
LGTM

PR #605, opened 2026-08-19
refactor(llms): bundle multi-argument params into structs; drop dead code

winterfx:
## Summary
- Fixed the two private, single-package argument-limit findings that were incorrectly deferred alongside the exported facade family they call into:
  - `resolveCustomOpenAIFacadeTarget` bundled into `customOpenAIFacadeTargetRequest`.
  - `ensureSessionOpenAIEnvProviderWithConfig` now takes the existing `SessionEnvProviderQuery` struct (already used by its Anthropic sibling) instead of inventing a new type — same shape, one less type.
  - `requireClaudeFacadeModel` (test helper) bundled into `claudeFacadeModelCheck`.
  
  Both non-test functions were originally deferred with reasoning copied from the exported family (`EnsureOpenAIEnvProvider` etc.) they call into, but being private and single-package themselves, they don't share that family's cross-package blast radius.

- Dead code found while auditing the package's exported surface (`ac-find-simplifications` pass, 116 funcs + 39 types) — all confirmed zero callers anywhere in the repo:
  - `EnsureSessionOpenAIEnvProvider` + its sole private caller `ensureSessionOpenAIEnvProvider` (dead forwarding wrapper).
  - `IsOptionalConfigError` (dead forwarding wrapper around `isOptionalConfigError`, which is still used internally).
  - `AnthropicProviderAuthFromLookup`.
  - `ManagedRuntimeEnvMap` — had a dedicated test assertion but zero production callers; removed the function and that assertion.

- **Deliberately not touched**: the exported `Ensure*FacadeConfig`/`Resolve*Target`/`NewFacadeToken` family — real cross-package callers in `pkg/agentcompose`, `pkg/runs`, `pkg/storage`. Spot-checked `ResolveRuntimeLLMTargetWithEnv`'s "36 call sites across 4 packages" nolint claim and confirmed it's accurate (14 files across those 4 packages), unlike the false-positive counts found in other packages this pass.
- Same as `pkg/driver`/`pkg/schedulers`/`pkg/sandboxes`/`pkg/runs`: `.golangci.yml` isn't included since other packages aren't done yet.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/llms/...`
- [x] `go vet ./pkg/llms/...`
- [x] Verified via `rg` that all 4 deleted symbols have zero references anywhere in the repo
- [x] Verified this branch builds and tests green in a clean isolated `git worktree` checkout of the commit (not just the local working tree, which has other uncommitted packages coexisting)

### DIFF
--- a/pkg/llms/agent_model_resolution.go
+++ b/pkg/llms/agent_model_resolution.go
@@ -98,10 +98,10 @@ func ResolveAgentModel(ctx context.Context, config *appconfig.Config, store Agen
 		return resolution, nil
 	}
 	if hasConfiguredProviderForFamily(providers, providerFamily) {
-		return selectConfiguredDefaultModel(ctx, store, providers, models, providerFamily)
+		return selectConfiguredDefaultModel(ctx, defaultModelSelection{Store: store, Providers: providers, Models: models, ProviderFamily: providerFamily})
 	}
 
-	resolution, ok, err = selectStoredDefaultModel(ctx, store, providers, models, providerFamily)
+	resolution, ok, err = selectStoredDefaultModel(ctx, defaultModelSelection{Store: store, Providers: providers, Models: models, ProviderFamily: providerFamily})
 	if err != nil {
 		return AgentModelResolution{}, err
 	}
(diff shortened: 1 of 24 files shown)


kingfs (approved):

PR #609, opened 2026-08-20
refactor(events): bundle BuildPayload params; split handleWebhook

winterfx:
## Summary
- Part of the funlen/argument-limit lint sweep across pkg/ (rule not yet enabled in `.golangci.yml`).
- `BuildPayload` (`pkg/events/webhooks/helpers.go`, exported): 8 positional params → `WebhookPayloadRequest` struct. Verified 3 real call sites across 2 packages (`cmd/agent-compose`, `pkg/events/webhooks`) — small enough blast radius to fix directly rather than defer.
- `handleWebhook` (`pkg/events/webhooks/http.go`, 107 lines) split into `resolveWebhookEvent` (validate, authorize, decode, idempotency check) and `storeWebhookEvent` (build payload, persist, respond) — two genuinely distinct phases already following the file's existing `handled bool, err error` early-return convention used by its sibling helpers.
- Left deferred (no `nolint`, since the rule isn't shipped in this PR): a coverage-sweep test, `ReleaseEventClaim`'s test fake (must match the `Store` interface's existing 6-arg signature, same as the already-deferred production implementation in pkg/storage), and two widely-used self-describing test helpers (`deliverGitHubWebhook`/`deliverGitHubWebhookWithContentType`, used across 10+ table-driven test call sites).

## Test plan
- [x] `go build ./...` (isolated `git worktree` at this branch's tip, generated `proto/` packages copied in since they're gitignored)
- [x] `go vet ./pkg/events/... ./cmd/agent-compose/...`
- [x] `go test ./pkg/events/... ./cmd/agent-compose/...`
- [x] `golangci-lint run ./pkg/events/...` — 4 remaining issues, all expected/deferred as described above

### DIFF
--- a/cmd/agent-compose/daemon_trusted_headers_webhook_test.go
+++ b/cmd/agent-compose/daemon_trusted_headers_webhook_test.go
@@ -16,10 +16,15 @@ func TestDaemonTrustedHeadersReachWebhookPayload(t *testing.T) {
 	app := echo.New()
 	app.Use(newDaemonTrustedHeadersMiddleware())
 	app.POST("/api/webhooks/:topic", func(c echo.Context) error {
-		payload := webhooks.BuildPayload(
-			c.Request(), "event-1", 1, c.Param("topic"), "correlation-1", "idempotency-1",
-			domain.WebhookSource{}, map[string]any{"intent": "push"},
-		)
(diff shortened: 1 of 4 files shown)


kingfs (approved):

PR #611, opened 2026-08-20
refactor(compose): split interpolateSkillSourceFields out of normalizeSkillSpec

winterfx:
## Summary
- Part of the funlen/argument-limit lint sweep across pkg/ (rule not yet enabled in `.golangci.yml`).
- `normalizeSkillSpec` (`pkg/compose/normalize.go`, 101 lines, barely over the limit): split out `interpolateSkillSourceFields`, which interpolates each raw `SkillSpec` field and assembles them into a normalized `sources.Source` — a distinct concern from the provider-specific validation that follows it.
- Behavior-preserving note: two `switch`-case comparisons that previously checked the raw interpolated `urlValue`/`pathValue` (`strings.TrimSpace(urlValue) == ""`, `strings.TrimSpace(pathValue) == ""`) were rewritten against the already-`.Normalized()` `commonSource.URL`/`.Path`. `Source.Normalized()` computes those exact fields as `strings.TrimSpace(...)` of the same raw values, so this is an equivalent read, not a behavior change — confirmed by the package's existing skill-normalization test suite passing unchanged.

## Test plan
- [x] `go build ./...` (isolated `git worktree` at this branch's tip, generated `proto/` packages copied in since they're gitignored)
- [x] `go vet ./pkg/compose/...`
- [x] `go test ./pkg/compose/...` — all skill-normalization tests pass; the one failure (`TestNormalizeCronTimezone`) is pre-existing on `main` and unrelated to this change (confirmed via `git stash` bisection against a clean checkout)
- [x] `golangci-lint run ./pkg/compose/...` — 0 issues

### DIFF
--- a/pkg/compose/normalize.go
+++ b/pkg/compose/normalize.go
@@ -233,7 +233,7 @@ func Normalize(spec *ProjectSpec, options NormalizeOptions) (*NormalizedProjectS
 			return nil, err
 		}
 		agent := spec.Agents[agentName]
-		normalizedAgent, err := normalizeAgent(agentName, agent, options, normalized.Volumes, normalized.Workspaces, normalized.MCPServers, normalized.OctoBusServers)
+		normalizedAgent, err := normalizeAgent(agentName, agent, options, normalized)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


kingfs (approved):

PR #664, opened 2026-09-02
fix(scheduler): stop collapsing agent finalText into the raw transcript

### LINKED ISSUE #663: bug(scheduler): webhook `finalText` is filled with the cell transcript, not the assistant's last message
## Summary

When a scheduler script returns `reply.finalText` (or `reply.text`) after `agent()`, the webhook payload contains the full cell transcript (thinking, shell commands, tool output), not the assistant's last user-facing message.

The runtime already distinguishes these two values. The scheduler host overwrites `FinalText` with `cell.Output`.

Still present on current `main` (`d1f78bd2`).

## Expected

| Field                            | Meaning                                        |
| -------------------------------- | ---------------------------------------------- |
| `reply.output`                   | full transcript (thinking / commands / stdout) |
| `reply.finalText` / `reply.text` | last assistant message shown to the user       |

A compose script like:

```js
text: reply.finalText || reply.text || reply.output
```

should prefer the final assistant message.

## Actual

`RuntimeHost.Agent()` copies `cell.Output` into `Text` and `FinalText`:

```go
// pkg/schedulers/run_host.go (~273)
finalText := firstHostNonEmpty(cell.Output, cell.Stdout, cell.Stderr)
result := domain.SchedulerAgentResult{
    Text:      finalText,
    Output:    cell.Output,
    FinalText: finalText, // same value as Output
}
```

So `finalText`, `text`, and `output` are the same string. Selecting `reply.finalText` has no effect.

Webhook consumers therefore receive the first/process text (e.g. "I'll search…", `cat SKILL.md`, raw JSON) instead of the later formatted answer.

This is not a missing webhook reply. `runtime.*.replied` is published; the wrong field is copied into it.

## Why this looks like a wiring bug

The lower layer already extracts the provider's last assistant message:

```go
// pkg/agentcompose/adapters/agent_executor.go (~314)
func agentAssistantMessage(result domain.AgentRunResult) string {
    if finalText := strings.TrimSpace(result.FinalText); finalText != "" &&
        result.FinalTextSource == domain.AgentFinalTextSourceProviderMessage {
        return finalText
    }
    ...
}
```

That value is stored on the `agent.assistant` event and returned from `ExecuteAgentRequest` (\~280 / \~299).

The scheduler adapter drops it:

```go
// pkg/agentcompose/adapters/scheduler_host.go (~40)
cell, _, _, err := e.Executor.ExecuteAgentRequest(...)
return cell, err
```

`NotebookCell` has `Output` (transcript) but no final-text field, so `run_host.go` can only fall back to the transcript.

`AgentFinalTextSource` already documents the intended split:

* `provider_message`: FinalText came from a provider assistant-message event

* `transcript_fallback`: compatibility fallback containing the transcript

## Suggested fix

1. Propagate the `agent.assistant` event from `ExecuteAgent` instead of discarding it with `_`.
2. If that event has a non-empty message, use it for `result.FinalText` and `result.Text`.
3. Keep `result.Output = cell.Output`.
4. If there is no assistant message, fall back to `cell.Output`.
5. Add a test: when an assistant message exists, `FinalText` != `Output`.

Do not try to recover the final answer by slicing `cell.Output` in the webhook consumer.

## Note

This only helps when the provider emits a distinct last assistant message. If a model puts thinking into that single final message, selecting last-message will not strip it; that is a separate model/prompt issue.

## Environment

* Repo: `chaitin/agent-compose`

* Branch: `main` @ `d1f78bd2`

* Trigger: scheduler event / webhook → `agent()` → `runtime.*.replied`

### bug

Image

### Expected

Image

winterfx:
## Summary
- Fixes #663: `scheduler.agent()` always set `text`/`finalText`/`output` to the same transcript string, so a script preferring `reply.finalText` got the raw process output (thinking, tool calls, JSON) instead of the provider's final message.
- `SchedulerHostAgentExecutor.ExecuteAgent` was discarding `ExecuteAgentRequest`'s assistant-message return value (`_, _`), leaving `RuntimeHost.Agent` (the bare-sandbox `scheduler.agent()` path) nothing but `cell.Output` to fall back on.
- Same bug, second occurrence: `AgentResultFromProjectRun` (used when the scheduler is bound to a managed project agent) only receives a `domain.ProjectRunRecord`, which never carried the assistant message at all — it echoed `run.Output` into `FinalText` too. Threaded the assistant message through `TransitionFromAgentCell`'s `ResultJSON` so this path can recover it.
- Both fixes keep `Output` untouched (still the full transcript) and only change what `Text`/`FinalText` resolve to.

## Test plan
- [x] `go test ./pkg/schedulers/... ./pkg/agentcompose/adapters/... ./pkg/runs/...` — new regression tests added:
  - `TestRuntimeHostAgentPrefersAssistantMessageOverTranscript`
  - `TestSchedulerHostAgentExecutorReturnsAssistantMessage`
  - `TestAgentResultFromProjectRunPrefersFinalTextMetadataOverOutput` / `...FallsBackToOutputWithoutFinalTextMetadata` / `...FallsBackToErrorWhenEmpty`
  - `TestTransitionFromAgentCellRecordsFinalTextMetadata`
- [x] Verified against a real running daemon (real Docker sandboxes, real LLM call via `chaitin/agent-compose-guest:latest` + configured `LLM_API_*`), exercising both `scheduler.agent()` paths through a manually-triggered scheduler script:
  - bare path (`RuntimeHost.Agent`): `reply.finalText` == `reply.text` == the clean provider answer, `reply.output` carries the raw transcript, `finalText !== output`.
  - project-agent path (`RuntimeHost.ProjectAgent` → `AgentResultFromProjectRun`): same result, and confirmed the new `finalText` field is actually persisted in the `project_run.result_json` DB column.

https://claude.ai/code/session_01QK1vvtPw374pVzpN8pG6AJ

### DIFF
--- a/pkg/agentcompose/adapters/adapter_helpers_coverage_test.go
+++ b/pkg/agentcompose/adapters/adapter_helpers_coverage_test.go
@@ -22,7 +22,7 @@ func TestAdapterHelperCoverage(t *testing.T) {
 		if _, err := (SchedulerHostEvents{}).AddRecord(context.Background(), schedulers.SchedulerEventInput{}); err == nil {
 			t.Fatalf("SchedulerHostEvents.AddRecord returned nil error")
 		}
-		if _, err := (SchedulerHostAgentExecutor{}).ExecuteAgent(context.Background(), nil, schedulers.HostAgentExecutionRequest{}); err == nil {
+		if _, _, err := (SchedulerHostAgentExecutor{}).ExecuteAgent(context.Background(), nil, schedulers.HostAgentExecutionRequest{}); err == nil {
 			t.Fatalf("SchedulerHostAgentExecutor.ExecuteAgent returned nil error")
 		}
 		if _, err := (SchedulerHostCommandExecutor{}).ExecuteSchedulerCommand(context.Background(), nil, domain.SchedulerCommandRequest{}); err == nil {
--- a/pkg/agentcompose/adapters/scheduler_host.go
+++ b/pkg/agentcompose/adapters/scheduler_host.go
@@ -33,11 +33,11 @@ type SchedulerHostAgentExecutor struct {
 	Executor *AgentExecutor
 }
 
-func (e SchedulerHostAgentExecutor) ExecuteAgent(ctx context.Context, session *domain.Sandbox, request schedulers.HostAgentExecutionRequest) (domain.NotebookCell, error) {
+func (e SchedulerHostAgentExecutor) ExecuteAgent(ctx context.Context, session *domain.Sandbox, request schedulers.HostAgentExecutionRequest) (domain.NotebookCell, string, error) {
 	if e.Executor == nil {
-		return domain.NotebookCell{}, fmt.Errorf("agent executor is unavailable")
+		return domain.NotebookCell{}, "", fmt.Errorf("agent executor is unavailable")
 	}
-	cell, _, _, err := e.Executor.ExecuteAgentRequest(ctx, session, execution.ExecuteAgentRequest{
+	cell, _, assistantEvent, err := e.Executor.ExecuteAgentRequest(ctx, session, execution.ExecuteAgentRequest{
 		Agent:             request.Provider,
 		AgentDefinitionID: request.AgentDefinitionID,
 		Model:             request.Model,
@@ -46,7 +46,7 @@ func (e SchedulerHostAgentExecutor) ExecuteAgent(ctx context.Context, session *d
 		Timeout:           request.Timeout,
 		OutputSchemaJSON:  request.OutputSchemaJSON,
 	})
-	return cell, err
+	return cell, assistantEvent.Message, err
 }
 
 type SchedulerHostCommandExecutor struct {
--- a/pkg/agentcompose/adapters/scheduler_host_test.go
+++ b/pkg/agentcompose/adapters/scheduler_host_test.go
@@ -1,0 +1,70 @@
+package adapters
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+	"time"
+
+	appconfig "github.com/chaitin/agent-compose/pkg/config"
+	driverpkg "github.com/chaitin/agent-compose/pkg/driver"
+	domain "github.com/chaitin/agent-compose/pkg/model"
+	"github.com/chaitin/agent-compose/pkg/schedulers"
+	"github.com/chaitin/agent-compose/pkg/storage/sandboxstore"
+)
+
+// TestSchedulerHostAgentExecutorReturnsAssistantMessage guards against
+// SchedulerHostAgentExecutor.ExecuteAgent silently discarding the provider's
+// final assistant message (as it did before this fix), which forced
+// RuntimeHost.Agent to fall back to the raw transcript for both Text and
+// FinalText.
+func TestSchedulerHostAgentExecutorReturnsAssistantMessage(t *testing.T) {
+	ctx := context.Background()
+	root := t.TempDir()
+	config := &appconfig.Config{
+		DataRoot:             root,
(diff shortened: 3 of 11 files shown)


eetoc:
### Code review

Found 1 issue:

1. On the agent failure path, `text`/`finalText` now surface a synthesized placeholder instead of the real transcript (bug due to `agentAssistantMessage` returning `"<agent> failed without output"`, and `persistFailedCell` setting `Message: fmt.Sprintf("%s run failed: %v", agent, finalErr)`)

The clearest signal is inside one of the changed files: two adjacent lines now treat the same `assistantEvent` with opposite levels of trust. Line 664 refuses to use its `Message` when the run failed, while line 663 (changed here) forwards it unconditionally:

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/pkg/runs/controller.go#L662-L665

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/pkg/runs/run_event_projection.go#L16-L22

On failure that `Message` is synthesized, never provider text. `persistFailedCell` is careful to substitute it into the output only when the transcript is empty:

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/pkg/agentcompose/adapters/agent_executor.go#L134-L142

Both changed fallback chains now rank that value ahead of the real transcript unconditionally, defeating that guard:

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/pkg/schedulers/run_host.go#L275-L277

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/pkg/schedulers/project_run_result.go#L18-L21

Verified locally: `AgentResultFromProjectRun` with `run.Output` = `"tool call failed: ENOENT /workspace/build.sh\nstack: ..."` and `ResultJSON.finalText` = `"codex run failed: exit status 1"` returns the synthesized string for **both** `Text` and `FinalText`; before this PR both returned the transcript. On the project-run path the transcript then has no route into `text`/`finalText` at all, since `metadata.FinalText` outranks `run.Output`. Every test added in this PR uses `Success: true`, so this path is uncovered.

This also crosses the documented contract, which scopes `finalText` to provider-sourced text:

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/docs/design/agent-compose-runtime_contract.md#L385-L389

Relatedly, the new doc comment states the message is `""` when the provider emitted nothing distinct from the transcript, which does not hold on the failure path:

https://github.com/chaitin/agent-compose/blob/c5d8b3289947538762bfbc33a8b7e6c36952dbdc/pkg/schedulers/run_host.go#L48-L52

Gating the preference on success — mirroring the `execErr != nil || !cell.Success` check already used on line 664 — would keep the intended fix while preserving failure diagnostics.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

winterfx:
Good catch, thanks — fixed in f9f35881.

Gated the finalText preference on the run having succeeded in both `RuntimeHost.Agent` (run_host.go) and `AgentResultFromProjectRun` (project_run_result.go), mirroring the `execErr != nil || !cell.Success` guard `run_event_projection.go` already uses for the same `assistantEvent`. On failure both paths now fall back to the real transcript exactly as before #664, and the synthesized placeholder from `agentAssistantMessage()` never displaces it.

Added regression tests for both paths using your exact repro shape (failed run, real diagnostic transcript in Output, synthesized summary as the would-be finalText).

eetoc (approved):

PR #667, opened 2026-09-04
fix(webhooks): /stop should also cascade to correlated events

winterfx:
## Summary
- `/api/webhooks/events/:event_id/stop` only walked the `parent_event_id` chain (`ListDescendantEventIDs`), so it never reached an event published as a follow-up webhook call (which never sets `parent_event_id`) even when it shares the same `correlation_id` as the event being stopped.
- Concretely: a router-style agent (e.g. `webhook.devboard.task.created`) that finishes and then POSTs a new webhook to kick off the actual long-running downstream agent (e.g. `webhook.agentkit.devboard.project.dev-board-server.task.created`) produces two sibling events with no parent/child link, only a shared `correlation_id`. Calling `/stop` on the router event was a no-op against the still-running downstream run.
- Added `Store.ListCorrelatedEventIDs` (implemented in `configstore.eventStore`) which returns every event sharing the root event's `correlation_id`, and merged it into the ID set `handleStopEvent` cancels/stops, alongside the existing parent-chain descendants.

## Note on scope
This intentionally does **not** restrict cascading to events published by the same webhook source — `correlation_id` is caller-supplied and, in the observed case, the router event and the downstream event actually come from two different registered webhook sources (`devboard` vs `agentkit-devboard-internal`) that are part of the same trusted internal flow. This was a deliberate simplification discussed with the reviewer; a source-scoped trust model was considered but intentionally deferred.

## Test plan
- [x] `go build ./pkg/... ./cmd/...`
- [x] `go test ./pkg/events/webhooks/... ./pkg/storage/configstore/...`
- [x] `go test ./pkg/agentcompose/... ./pkg/events/... ./pkg/storage/... ./pkg/schedulers/...`
- [x] Unit tests in `event_stop_test.go`: a correlated event outside the parent chain is stopped too; a correlated event already in the parent chain is not duplicated.
- [x] **New real end-to-end regression test** (`TestIntegrationWebhookStopCascadesToCorrelatedEventAcrossSources` in `pkg/agentcompose/app/webhook_stop_integration_test.go`): spins up a real sqlite-backed store + real scheduler controller with two projects/schedulers, dispatches two real (execution-blocked) runs from two events that share a `correlation_id` but come from two different webhook sources and have no `parent_event_id` link — reproducing the exact production shape. Calling `/stop` with only the router source's token cancels both runs.
  - Verified locally that this test **fails** (`requested_runs=1`, the correlated worker run stays running) when the `ListCorrelatedEventIDs` merge is disabled, confirming it actually exercises the fix rather than passing vacuously.

https://claude.ai/code/session_01SaFrW4iFGfXYUredLVPxiH

### DIFF
--- a/docs/design/webhook_design.md
+++ b/docs/design/webhook_design.md
@@ -549,6 +549,24 @@ main implementation path for a formal API:
 - Therefore `GET /api/events/:event_id/sandboxes` should not rely on full-table
   JSON scan as the main implementation.
 
+The exact-relation restriction above governs exact attribution in
+relation-oriented read models. It does not define the cancellation scope of
+`POST /api/webhooks/events/:event_id/stop`. That endpoint deliberately treats
+the requested event's `correlation_id` as a group-wide cancellation domain and
+cancels the union of:
+
+- the requested event and its descendants found through `parent_event_id`; and
+- every stored event that shares the requested event's `correlation_id`.
+
+The correlation group has no causal direction, so this cancellation scope can
+include earlier, later, and sibling events, including events submitted through
+different webhook sources. The endpoint authenticates the static token and
+enabled state of the requested event's webhook source only; it does not
+separately authenticate or require an enabled source for every correlated
+event. This is an explicit control-plane policy, not a claim that
+`correlation_id` establishes an exact causal relation, and it must not be reused
+as exact attribution by relation-oriented APIs.
+
 Formal implementation needs explicit relation tables. Event-to-run relation is
 stored in `event_delivery`; event-to-sandbox relation is stored in
 `event_sandbox_link`:
@@ -576,9 +594,11 @@ Write timing:
   `event_delivery(event_id, scheduler_id, trigger_id, scheduler_run_id, status=run_started)`.
 - When the scheduler run host sees `linked_sandbox_id`, also write
   `event_id -> sandbox_id`.
-- When a scheduler derives events, write `parent_event_id`. Query
-  `GET /api/events/:event_id/sandboxes` should expand descendant events by
-  `parent_event_id`, then aggregate `event_sandbox_link` for those events.
+- When a scheduler derives events, write `parent_event_id`. For exact sandbox
+  attribution, `GET /api/events/:event_id/sandboxes` should expand descendant
+  events by `parent_event_id`, then aggregate `event_sandbox_link` for those
+  events. This causal expansion rule does not limit the correlation-wide
+  cancellation scope defined above.
 - If business wants the original webhook event to find sandboxes created by
   derived events, query layer expands descendants; descendants do not need to
   write duplicate ancestor links.
--- a/docs/pages/command-line-manual.md
+++ b/docs/pages/command-line-manual.md
@@ -127,7 +127,11 @@ authentication is enabled.
 #### Stopping webhook-triggered runs
 
 The client that submitted a webhook can later request cancellation of every
-active run associated with that event and its descendant events:
+active run associated with that event, its descendant events, and every event
+that shares the requested event's `correlation_id`. Correlation scope is
+group-wide rather than descendant-only: it includes earlier, later, and sibling
+events, even when those correlated events were submitted through different
+webhook sources.
 
 ```http
 POST /api/webhooks/events/<event-id>/stop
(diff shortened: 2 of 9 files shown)


eetoc:
### Code review

Found 3 issues:

1. Cross-source authorization bypass: `/stop` authenticates only against the **root** event's `webhookSourceId`, then cancels every correlated event's runs with no per-event re-authorization. `correlation_id` is untrusted caller input (`ExtractCorrelationID` reads `X-Correlation-ID` or the body), whereas `webhookSourceId` is server-set from the authenticated source — so this keys a destructive cascade on the forgeable field. A holder of any one source's token can publish an event with a colliding `correlation_id` and stop another source's runs. This also re-opens finding `e3631ae4` (fixed in 83b9d610): only the root's `source.Enabled` is checked, so a **disabled** source's runs can still be stopped via a correlated event. PR #649's review signed off the endpoint on the opposite invariant — "只有事件所属 source 的 token 持有者才能取消其级联运行，未发现越权或伪造路径".

https://github.com/chaitin/agent-compose/blob/46dce31fe1bf3f33e3eb7ff5e08c6c27565f1289/pkg/events/webhooks/event_stop.go#L81-L87

2. The cascade has no causal direction, so `/stop` on a late event also cancels its ancestors and unrelated siblings. The query selects the whole correlation group with no descendant restriction, turning a leaf stop into "stop the entire correlation group". The design doc rules this out directly — `correlation_id` "may cover multiple derived events and runs; it is only a trace helper and cannot be used alone as an exact relation" ([webhook_design.md#L547-L548](https://github.com/chaitin/agent-compose/blob/46dce31fe1bf3f33e3eb7ff5e08c6c27565f1289/docs/design/webhook_design.md#L546-L549)). The same objection was raised as a blocking review on #573, which applied this merge only to the read-only trace path. The new tests cover only the forward direction.

https://github.com/chaitin/agent-compose/blob/46dce31fe1bf3f33e3eb7ff5e08c6c27565f1289/pkg/storage/configstore/topic_event_store.go#L542-L544

3. The `/stop` API contract changed but no docs were updated (AGENTS.md says "Update docs for behavior, configuration, API, or workflow changes"). The manual still documents the old descendant-only scope; precedent 144b1ed5 updated both this file and its zh-CN copy when it changed `/stop` semantics.

https://github.com/chaitin/agent-compose/blob/46dce31fe1bf3f33e3eb7ff5e08c6c27565f1289/docs/pages/command-line-manual.md#L128-L131

Also worth a look (not blocking): `mergeCorrelationEventIDs` in [event_query.go#L240](https://github.com/chaitin/agent-compose/blob/46dce31fe1bf3f33e3eb7ff5e08c6c27565f1289/pkg/storage/configstore/event_query.go#L239-L242) already solves this same problem, budget-aware and with a truncation flag; the new `ListCorrelatedEventIDs` + `mergeEventIDs` reimplements it less carefully.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

eetoc:
### Code review (update for 8dbb9ad)

Re-reviewed the delta since 46dce31 — one commit, docs only, no Go changes.

**Issue 3 (docs) is resolved.** Both manual copies now describe the group-wide scope accurately and stay in sync, and they state the authentication boundary plainly rather than glossing it.

**Issues 1 and 2 are documented, not fixed.** The cascade code is byte-identical, so a token for one source still cancels another source's runs, and a disabled source's runs are still reachable through a correlated event. What changed is that the manual now writes that gap down as the contract:

https://github.com/chaitin/agent-compose/blob/8dbb9ad8818ef2464251f8627f25ab7c12e940ff/docs/pages/command-line-manual.md#L144-L148

Documenting a cross-source authorization gap makes it intentional, not safe — worth an explicit sign-off from whoever owns the trust model, since #649 was approved on the opposite invariant.

One new issue:

1. The design doc now contradicts the manual and was not updated. It still states `correlation_id` "is only a trace helper and cannot be used alone as an exact relation", and specifies that causal expansion goes through `parent_event_id` — while the manual now documents correlation-group expansion as the `/stop` contract. Since this is the constraint #573 was blocked on, leaving both statements in the tree means the next reader gets opposite answers depending on which file they open.

https://github.com/chaitin/agent-compose/blob/8dbb9ad8818ef2464251f8627f25ab7c12e940ff/docs/design/webhook_design.md#L546-L549

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

eetoc:
### Code review (update for ba53156)

Re-reviewed the delta since 8dbb9ad — one commit, design doc only.

**The doc contradiction I raised is resolved, and resolved well.** Rather than deleting the old constraint, the new section separates the two domains: exact attribution in relation-oriented read models still may not lean on `correlation_id`, while `/stop` treats the correlation group as a cancellation domain by explicit policy. It also states the authentication and enabled-state boundary directly, and adds a guardrail against reuse.

One issue with that guardrail:

1. The new clause says the correlation-wide scope "must not be reused as exact attribution by relation-oriented APIs", but `GET /api/events/:event_id/trace` already does exactly that. `GetEventTrace` merges correlation siblings into `eventIDs` via `mergeCorrelationEventIDs`, then aggregates `listEventRunTraces` and `ListEventSandboxLinks` over that widened set — so another event's runs and sandbox links are attributed to the requested event's trace. That was added in #573, the PR blocked on this same constraint. As written the doc names `/stop` as the only exception while a second one is already shipped, so the guardrail reads as prohibiting existing behavior. Worth either acknowledging the trace path as in-scope policy too, or stating why its expansion is not "exact attribution".

https://github.com/chaitin/agent-compose/blob/ba531568f277b23bdb7b7498b85e759cbaae542b/docs/design/webhook_design.md#L565-L569

For the record on the earlier findings: no Go file has changed between 46dce31 and ba53156, so the cross-source cancellation and the directionless correlation scope are unchanged in behavior. They are now stated policy in both the manual and the design doc rather than undocumented behavior, which is the right way to land them if the team accepts the trust model — that acceptance is the remaining decision, not a code question.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

eetoc (approved):
LGTM

PR #685, opened 2026-09-10
feat: apply LLM_API_HEADERS to default env LLM providers

Heisenhower:
## Problem

Default/session env providers currently persist `HeadersJSON: "{}"`, so daemon `LLM_*` config cannot carry static gateway headers. Catalog providers already support headers, but complete daemon env config outranks catalog.

This blocks products that configure a default LLM through GlobalEnv (for example an OpenAI-compatible gateway that requires extra headers such as `Bizscenario`) from sending those headers on `LLMService.Generate` and the runtime LLM facade.

## Solution

Read `LLM_API_HEADERS` (JSON object of static `string -> string` headers) during OpenAI and Anthropic env bootstrap and store it on the existing provider `HeadersJSON`.

- Empty / missing / `{}` means no extra headers.
- Anthropic env providers still include `anthropic-version` and merge custom headers on top.
- Invalid JSON, non-object values, empty names, or CR/LF fail bootstrap/resolve instead of sending a request with unknown headers.
- Forwarding still uses `ProviderForwardHeaders`; `Authorization` and other system headers continue to come from the API key.

Out of scope: catalog UI, dynamic/signed headers, proto changes.

## Tests

- `go test ./pkg/llms/`


eetoc:
Thanks for this — the gap is real. An env-configured default provider genuinely could not carry gateway headers before, and the problem statement names the case precisely. A few things I want to note as done right before the suggestions: `Authorization` really is protected in depth (custom headers are filtered by `ForbiddenProviderHeader` *and* the auth header is set last from the API key, so neither path can override it), `.env.example` and all three doc pages were updated together, and the stated principle — fail at bootstrap rather than send an unknown header — is the right one.

Five suggestions below, roughly in order of how much I'd want them addressed. Only the first one I'd ask for before merge; the rest are cheap enough to fold in while you're here.

## 1. `LLM_API_HEADERS` isn't on the guest-runtime credential denylist

`LLMProviderKeyName` (`pkg/driver/types.go:198`) is documented as the canonical denylist of credentials that "must never be passed through to a guest runtime". It lists eight names, all `*_API_KEY` / `*_AUTH_TOKEN` shaped, and `LLM_API_HEADERS` is not among them.

The path it takes: `pkg/runs/preparation.go:86` merges global env into the sandbox's `EnvItems`; `FilterPersistedRuntimeEnv` (`pkg/llms/env.go:48`) strips only those eight plus `AGENT_COMPOSE_RUNTIME_BASE_URL`; `sandboxEnvMap` (`pkg/driver/types.go:183`) applies the same list. So the value is persisted on the sandbox and injected into the guest, where any agent or code it runs can read it.

What makes this worth closing is the contrast with its sibling: `LLM_API_KEY` *is* stripped and replaced with a short-lived facade token (`pkg/llms/runtimefacade/config.go:136-144`). For a gateway that authenticates with a header token, `LLM_API_HEADERS` carries the other half of the same credential — so today the boundary is enforced for one half and not the other. And since `LLM_API_ENDPOINT` is not stripped either, the sandbox would hold both the upstream address and a long-lived credential for it, bypassing the facade entirely.

**This does not conflict with the feature.** I checked every reference on this branch: the only consumer is `pkg/llms/env_headers.go`, daemon-side. There are zero references under `runtime/`, `assets/` or `guest-images/`, and `Dockerfile.agent-compose-guest` installs only `agent-compose-runtime` plus the provider CLIs — the `agent-compose` Go binary, and therefore this code, is not in the guest image at all. The headers still reach upstream because the daemon applies them in `ProviderForwardHeaders` when it builds the `ResolvedTarget` (`pkg/llms/target.go:41`). Adding the name to `LLMProviderKeyName` costs the feature nothing; it just stops a value nothing in the sandbox reads from being handed to it.

(If I've misread and some scenario does need the agent to see these headers, please say so — that would change the recommendation.)

## 2. Header merge is case-sensitive, so a differently-cased override is non-deterministic

`envProviderHeadersJSON` keys the map by the literal JSON spelling, while HTTP field names are case-insensitive (RFC 9110 §5.1) and `ProviderForwardHeaders` applies them with `headers.Set`, which canonicalises.

Setting `LLM_API_HEADERS={"Anthropic-Version":"2024-01-01"}` therefore persists *both* entries:

```
{"Anthropic-Version":"2024-01-01","anthropic-version":"2023-06-01"}
```

and at request time the two canonicalise to the same header, so the last one visited wins — which Go randomises per range. Measured over 200 calls on this branch:

```
Anthropic-Version → {2023-06-01: 173, 2024-01-01: 27}
```

Same config, same process, the header alternates. Worth fixing less because someone is likely to hit it — the motivating headers (`Bizscenario`, `Branchid`) collide with nothing — and more because the response to any input should be either deterministic or an error, and this is neither. Canonicalising the key on insert (`textproto.CanonicalMIMEHeaderKey`) is enough.

Also worth noting: the existing tests use the lowercase spelling on both sides, which is why this path is uncovered — the same spelling is naturally reused when the test and the implementation are written together.

## 3. Validation rejects CR/LF but not other invalid name characters

The PR's contract is that bad input fails bootstrap rather than reaching a request, but only `\r`/`\n` and empty names are checked. `{"Biz Scenario":"mobile-learning"}` — a plausible typo for your own `Bizscenario` example — passes bootstrap and then fails *every* LLM call at the transport with `net/http: invalid header field name`. `httpguts.ValidHeaderFieldName` / `ValidHeaderFieldValue` would move that back to bootstrap where you intended it.

## 4. One headers variable, two independent upstreams

Endpoint, key and model each have an Anthropic-specific override (`ANTHROPIC_BASE_URL`, `ANTHROPIC_API_KEY`, `ANTHROPIC_MODEL`), which lets the Anthropic provider point at a completely different upstream. Headers have only the one variable, and `ensureAnthropicEnvProvider` reads it from the same lookup. So a deployment with `LLM_API_ENDPOINT` + `LLM_API_HEADERS` on a corporate gateway and `ANTHROPIC_BASE_URL` on `api.anthropic.com` would send the gateway's headers — including a token, if it has one — to Anthropic. Either scoping the variable to the family that owns it or adding an `ANTHROPIC_API_HEADERS` override would resolve it; documenting the fallthrough would at least make it a choice.

## 5. Two smaller ones

- The key was correctly added to the `isolateLLMEnv` helpers in `pkg/llms` and `pkg/llms/runtimefacade`, but `pkg/storage/configstore/llm_catalog_store_test.go:296` and `pkg/agentcompose/app/project_agent_model_resolver_test.go:18` clear the same `LLM_*` set and also exercise env-provider bootstrap. A developer who follows the new `.env.example` and exports the variable would see local-only failures there.
- The four new `fmt.Errorf` values aren't classified, while comparable configuration failures on this path use `domain.ClassifyError(domain.ErrFailedPrecondition, ...)` (e.g. `pkg/llms/resolver.go:400`). As written, a stray comma in the JSON surfaces as a generic internal error rather than a 400-class message naming the variable — which is the diagnosability this validation was added for.

## One process note

This is a fork PR, so no workflow has run yet — lint, Go tests, and the proto freshness check are all still pending a maintainer approval. Worth getting them green before merge, since `go test ./pkg/llms/` is currently the only verification the change has had.

eetoc (approved):

